PR #1, opened 2024-08-27
fix: issue_comments linting added issue_comments:edited, created and …

### LINKED ISSUE #32: Contextual Search Command
Perhaps as a stepping stone to the generalize, context aware, @UbiquityOS: question syntax, we can make a dedicated slash command and then merge the interface later. This allows the opportunity to focus on the logic instead of the interface in the near future. 

The problem is that contributors need to sometimes search for specific comments (or issue specification in this case) to add value to an existing thread. 

Parse the conversation and then hunt down what is needed. The slash command should be temporarily until we can merge it into the @UbiquityOS natural syntax. 

/search 

> There was a task carried out about this I remember, let me find it.
https://github.com/ubiquity/ubiquibot-logger/pull/35

_Originally posted by @gentlementlegen in https://github.com/ubiquity-os-marketplace/text-conversation-rewards/pull/132#discussion_r1787158520_

sshivaditya2019:
Resolves [#32](https://github.com/ubiquibot/plugins-wishlist/issues/32)

- Created webhook listener for issue_comments:created and issue_comments:edited.
- Uses OpenAI embeddings.
- Embeddings are stored in Supabase DB
- New OpenAI adapter can be expanded in future to support text generation and question answering using RAG.
- issue_comments:deleted, deletes the comment from the DB
- Expects the`Comment ID` , `Issue Body` in the payload.
- OpenAI Key, Supabase Key and URL have to be added in the envs.
- When deployed a new table has to be created with the associated supabase account with the following schema. For Vector type, the pgvector extension must be enabled in the postgres extensions.
```
issue_comment
id: int8
issuebody: Text,
commentbody: Text,
embedding: Vector(3072)
```


sshivaditya2019:
@0x4007

.vscode/settings.json
0x4007, line 1: Delete

package.json
```diff
@@ -17,7 +17,7 @@
     "knip-ci": "knip --no-exit-code --reporter json --config .github/knip.ts",
     "prepare": "husky install",
     "test": "jest --setupFiles dotenv/config --coverage",
-    "worker": "wrangler dev --env dev --port 4000"
+    "worker": "wrangler dev --env dev --port 5000"
```
0x4007, line 20: Why?

src/adapters/openai/helpers/embedding.ts
```diff
@@ -0,0 +1,25 @@
+import OpenAI from "openai";
+import { Context } from "../../../types";
+import { SuperOpenAi } from "./openai";
+const VECTOR_SIZE = 512;
```
0x4007, line 4: Why?

src/adapters/openai/helpers/embedding.ts
```diff
@@ -0,0 +1,25 @@
...
+  protected context: Context;
+
+  constructor(client: OpenAI, context: Context) {
+    super(client, context);
+    this.context = context;
+  }
+
+  async createEmbedding(text: string): Promise<number[]> {
+    const params: OpenAI.EmbeddingCreateParams = {
+      model: "text-embedding-3-small",
```
0x4007, line 16: Is this the best model?

src/main.ts
```diff
@@ -10,14 +10,13 @@ import { plugin } from "./plugin";
...
   const payload = github.context.payload.inputs;
-
   const env = Value.Decode(envSchema, payload.env);
   const settings = Value.Decode(pluginSettingsSchema, Value.Default(pluginSettingsSchema, JSON.parse(payload.settings)));
 
   if (!pluginSettingsValidator.test(settings)) {
     throw new Error("Invalid settings provided");
   }
-
+  console.log("ENV", payload);
```
0x4007, line 19: Maybe remove before we merge 
```suggestion
```

src/plugin.ts
```diff
@@ -2,47 +2,52 @@ import { Octokit } from "@octokit/rest";
...
 
 /**
  * How a worker executes the plugin.
  */
 export async function plugin(inputs: PluginInputs, env: Env) {
   const octokit = new Octokit({ auth: inputs.authToken });
-
+  const supabase = createClient(env.SUPABASE_URL, env.SUPABASE_KEY);
+  const openaiClient = new OpenAI({
+    apiKey: env.OPENAI_API_KEY,
```
0x4007, line 40: Prefer to check and throw error if key is empty first.

src/types/env.ts
```diff
@@ -10,7 +10,11 @@ import { StandardValidator } from "typebox-validators";
  * taken from either `dev.vars` or repository secrets.
  * They are used with `process.env` but are type-safe.
  */
-export const envSchema = T.Object({});
+export const envSchema = T.Object({
+  SUPABASE_URL: T.String(),
+  SUPABASE_KEY: T.String(),
+  OPENAI_API_KEY: T.String(),
```
0x4007, line 16: Oh perhaps this throws the error if it's empty

tests/main.test.ts
```diff
@@ -141,6 +145,7 @@ function createContextInner(
 ): Context {
   return {
     eventName: "issue_comment.created",
+    //@ts-expect-error modified for testing purposes
```
0x4007, line 148: @gentlementlegen rfc

0x4007:
Fix resolves by placing the full URL of the issue there.

src/plugin.ts
sshivaditya2019: We are validating the schema before running the plugin. I think this would be redundant

src/adapters/openai/helpers/embedding.ts
sshivaditya2019: image

Not the best model, but gets the job done. Should be good enough for smaller length messages and lower dim vectors.

package.json
sshivaditya2019: I am trying to replicate this, but I am not able to launch wrangler on port 4000.

Platform:
Wrangler@3.62.0
Node 20.11.1
MacOS

tests/main.test.ts
gentlementlegen: You can also cast the object to the proper type, I believe it gives better auto-complete.

src/adapters/openai/helpers/embedding.ts
sshivaditya2019: Could be expanded to add support, for text generation and question answering. I think it should be kept as a separate adapter.

package.json
0x4007: @gentlementlegen rfc

src/adapters/openai/helpers/embedding.ts
sshivaditya2019: - Vector search with large model could be slow with large number of entries.
- The larger dimensions size could be useful if we are designing a RAG based system, but for other purposes a smaller dimension size of around 512 to 1024 should be enough.
- The data chunks are smaller, so we could get away with a smaller vector dim.

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "plugin-template",
+  "name": "plugin-issue-comment",
```
gentlementlegen, line 2:
```suggestion
  "name": "@ubiquibot/issue-comment-embeddings",
```

package.json
gentlementlegen: Don't you have anything running already on port 4000? What error do you get?

sshivaditya2019: No, I don't have anything running on port 4000. 

Possible bug in wrangler [Wrangler Issue](https://github.com/cloudflare/workers-sdk/issues/4709)

src/adapters/openai/helpers/embedding.ts
0x4007: I think we should optimize as needed instead of proactively. This is because we are just assuming that performance may some day be a problem. 

Right now I consider this more of a research project, and honestly have basically no experience building these sort of systems.

sshivaditya2019: Just to be clear, this is a design issue as well. Depending on the type of model we choose we need to change the datatype in the vector store. Increasing the vector embedding dimension, after the vector store has been populated would be a little difficult.

And, I do agree with what you're saying, but performance and cost aside, these models perform the more or less the same on MTEB benchmark as well. I usually try to start with smaller models, and work my way up, if the application requires it.

gentlementlegen:
@sshivaditya2019 It seems that you are storing the vectors within Supabase but I see no Supabase folder for setup. Also, typings should be used so Supabase offers auto-complete.

src/adapters/openai/helpers/embedding.ts
sshivaditya2019: Fixed in the latest commit, uses large model.

sshivaditya2019:
> @sshivaditya2019 It seems that you are storing the vectors within Supabase but I see no Supabase folder for setup. Also, typings should be used so Supabase offers auto-complete. Please fix the unit tests as well.

Fixed added schema for table. Unit tests work now. Schema typing should be available in the project. Embedding type now expects a vector array of length 3072

wrangler.toml
```diff
@@ -1,4 +1,4 @@
-name = "your-plugin-name"
+name = "issue-comment-embeddings"
```
gentlementlegen, line 1:
```suggestion
name = "ubiquibot-issue-comment-embeddings"
```

0x4007: I think we should prefix all with `uos-` because ubiquibot is deprecated.

gentlementlegen:
@sshivaditya2019 Thanks a lot for the changes, appreciated. Since you seem to use Supabase, I assume that we should spin up a new instance specially for this project because this data would not make much sense in our current database, @0x4007 please confirm.

As such could you please add a migration script that would:
- run migrations
- update database types

when code is pushed to the main branch? Here is an example to get you started: https://github.com/ubiquibot/user-activity-watcher/blob/development/.github/workflows/database.yml#L35
It would be also nice to have code generation commands within the `package.json`, I don't think I saw them. Here is an example: https://github.com/ubiquibot/user-activity-watcher/blob/development/package.json#L21

0x4007:
Sure new database is fine to keep things simple.

wrangler.toml
0x4007: I wonder if it makes sense to also rename the repositories to match.

sshivaditya2019:
> @sshivaditya2019 Thanks a lot for the changes, appreciated. Since you seem to use Supabase, I assume that we should spin up a new instance specially for this project because this data would not make much sense in our current database, @0x4007 please confirm.
> 
> As such could you please add a migration script that would:
> 
> * run migrations
> * update database types
> 
> when code is pushed to the main branch? Here is an example to get you started: https://github.com/ubiquibot/user-activity-watcher/blob/development/.github/workflows/database.yml#L35 It would be also nice to have code generation commands within the `package.json`, I don't think I saw them. Here is an example: https://github.com/ubiquibot/user-activity-watcher/blob/development/package.json#L21

Added Github Workflow, and fixed the package.json.

src/adapters/openai/helpers/embedding.ts
```diff
@@ -0,0 +1,25 @@
+import OpenAI from "openai";
+import { Context } from "../../../types";
+import { SuperOpenAi } from "./openai";
+const VECTOR_SIZE = 3072;
```
gentlementlegen, line 4: We could consider having this parameter in the configuration later for easy changes and testing.

sshivaditya2019: I could add this in env for now. Is there a global config file ?

tests/__mocks__/adapter.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export interface CommentMock {
+  id: number;
+  commentbody: string;
+  issuebody: string;
+  embedding: number[];
+}
+
+export function createMockAdapters(context: Context) {
+  const commentMap: Map<number, CommentMock> = new Map();
```
gentlementlegen, line 14: I guess this would be reset every run. If we run twice the logic on the same group of comments, would this be problematic?

sshivaditya2019: Isn't that the expected behavior ? In case of collision it should throw error. I don't think there would be collision in comment id.

gentlementlegen: During the run it is fine, but all the results are pushed to supabase correct? I just wondered the impact of having duplicate content in the database.

sshivaditya2019: While its running, the `issue_comment.created` will create a new comment in the db. All the results are pushed to the database assuming there would be no collision in the `comment id`. Comments may contain duplicate content, but they cannot have the same ID.

gentlementlegen:
Please fix the Knip issues, I will try to set up this on my repo and give it a run.

sshivaditya2019:
Knip should be fixed in the latest commit.

src/types/database.ts
```diff
@@ -1,3 +1,5 @@
+// cSpell:disable
```
gentlementlegen: Since this is auto-generated I think you'd be better setting the ignore of this file through the configuration.

gentlementlegen (commented):
I had a horrendous time trying this locally and had to rely on a GitHub deployment to test it. A few major changes to consider:
- remove the `commands` section from the manifest otherwise the plugin will never run
- inside `compute.yml` rename the content properly (`plugin-name`)
- in `compute.yml` add the GitHub secrets properly or the run will crash (I assume this will run as an Action because it's a long running task)

Related run: https://github.com/Meniole/issue-comment-embeddings/actions/runs/10612546023/job/29414425736#step:5:8

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ts-template",
   "description": "ts-template for Ubiquibot plugins.",
```
gentlementlegen, line 3: Please change `name` and `description` properly.

sshivaditya2019:
> I had a horrendous time trying this locally and had to rely on a GitHub deployment to test it. A few major changes to consider:
> 
> * remove the `commands` section from the manifest otherwise the plugin will never run
> * inside `compute.yml` rename the content properly (`plugin-name`)
> * in `compute.yml` add the GitHub secrets properly or the run will crash (I assume this will run as an Action because it's a long running task)
> * update `README.md` properly to help future collaborators
> 
> Related run: https://github.com/Meniole/issue-comment-embeddings/actions/runs/10612546023/job/29414425736#step:5:8

- I expected this to run on worker, since action takes too long for each message.
- Its working for me as expected locally, might be some issue in `.ubiquibot-config.yml`. Could you please let me know what issue you were facing?
- Updated the`README.md`

gentlementlegen (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "plugin-template",
+  "name": "@ubiquibot/issue-comment-embeddings",
   "version": "1.0.0",
-  "description": "Ubiquibot plugin template repository with TypeScript support.",
+  "description": "Issue comment plugin for Ubiquibot.",
```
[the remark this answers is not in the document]
sshivaditya2019: @gentlementlegen How about `Worker plugin for generating vector embeddings of issue comments and storing them in a database` ?

tests/__mocks__/helpers.ts
```diff
@@ -37,7 +37,7 @@ export async function setupTests() {
     labels: [],
   });
 
-  createComment("/Hello", 1);
+  createComment("Hello, World!", 1);
```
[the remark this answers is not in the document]
sshivaditya2019: It was there for `createContext`. There was comment that mentioned to `create it first then pull it from the DB and feed it to _createContext`. Removed that now.

wrangler.toml
```diff
@@ -1,7 +1,8 @@
-name = "your-plugin-name"
+name = "ubiquibot-issue-comment-embeddings"
 main = "src/worker.ts"
 compatibility_date = "2024-05-23"
 node_compat = true
 
 [env.dev]
+PLUGIN_CHAIN_STATE = "4f7aadc56bef41a7ae2cc8c0582320b3"
```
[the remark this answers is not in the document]
sshivaditya2019: Removed that. I was under the impression kernel expects that value.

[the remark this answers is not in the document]
gentlementlegen: The kernel uses it to track the status of the plugins, but within the plugins this is not needed.

src/main.ts
```diff
@@ -10,7 +10,6 @@ import { plugin } from "./plugin";
  */
 export async function run() {
   const payload = github.context.payload.inputs;
-
   const env = Value.Decode(envSchema, payload.env);
```
gentlementlegen, line 13: This will always crash on run:
https://github.com/Meniole/issue-comment-embeddings/actions/runs/10628609152/job/29463778679#step:5:45

The environment is populated either from GitHub or the Worker. Since this runs as an action, it should be loaded directly within `process.env`, which you can add inside `payload.env` if you want. In the case of a worker, it is send within the payload by Cloudflare.

gentlementlegen (changes requested):
Some more changes might be needed to be able to run this one, but getting closer (c.f. my last comment).

---

Testing actions is tedious. I would advise you to create your own organization (also works on your personal account but usage restrictions are way lower) and set up the bot so you can try some runs and make sure it works as intended.

src/main.ts
sshivaditya2019: But, as mentioned before, I expect this to run in a Cloudflare worker environment. The env variables are sent within the payload.

sshivaditya2019:
Could you please confirm if the .ubiquibot-config.yml file isn't working for you when you try using it? I tested it again by creating a new issue and adding comments, and everything worked fine on my end. If needed, I can share a screencast to demonstrate this.

Is there any particular reason you keep trying to set it up in Github Action environment. The plugin could be made for a Cloudflare Worker or Standalone Github action right ? I don't want this to run in Github Action environment and is intended to be setup on Worker env.

sshivaditya2019:
https://github.com/user-attachments/assets/1af6065a-8ed0-4a61-a022-ee36b41f9513

0x4007:
Very lovely QA testing video thank you.

.github/.ubiquibot-config.yml
```diff
@@ -0,0 +1,6 @@
+plugins:
+  - name: test-app
+    id: test-app
+    uses:
+      - plugin: https://ubiquibot-issue-comment-embeddings.sshivaditya.workers.dev
+        runsOn: [ "issue_comment.created", "issue_comment.edited", "issue_comment.deleted" ]
```
0x4007, line 6: Can't we do `"issue_comment"` to be concise? @gentlementlegen

manifest.json
```diff
@@ -1,11 +1,5 @@
...
-  "name": "ts-template",
-  "description": "ts-template for Ubiquibot plugins.",
-  "ubiquity:listeners": ["issue_comment.created"],
-  "commands": {
-    "command1": {
-      "ubiquity:example": "/command1 argument",
-      "description": "Command 1 with an argument."
-    }
-  }
+  "name": "@ubiquibot/issue-comment-embeddings",
```
0x4007, line 2: ubiquity-os

manifest.json
```diff
@@ -1,11 +1,5 @@
...
+  "description": "Issue comment plugin for Ubiquibot. It enables the storage, updating, and deletion of issue comment embeddings.",
+  "ubiquity:listeners": ["issue_comment.created", "issue_comment.edited", "issue_comment.deleted"]
```
0x4007, line 4: Possibly `issue_comment` only

src/main.ts
0x4007: I think running as worker is fine.

wrangler.toml
```diff
@@ -1,4 +1,4 @@
-name = "your-plugin-name"
+name = "ubiquibot-issue-comment-embeddings"
```
0x4007, line 1: UbiquiBot term is deprecated in favor of UbiquityOS (or when lowercase, ubiquity-os)

0x4007 (approved):
This pull looks great. I just have some cosmetic changes

.github/.ubiquibot-config.yml
gentlementlegen: Not with the current implementation because it wouldn't match the event and wouldn't be triggered.

gentlementlegen:
@0x4007 My concern about having this as a worker is the following:
- we don't need to handle commands so I don't see the need for instant feedback
- workers are limited to 50 network calls and this plugin will most likely hit the DB a lot (each supabase action is a network call)
- it's a long running process

But I will test it as a worker and see if it works.

sshivaditya2019:
> @0x4007 My concern about having this as a worker is the following:
> 
> - we don't need to handle commands so I don't see the need for instant feedback
> 
> - workers are limited to 50 network calls and this plugin will most likely hit the DB a lot (each supabase action is a network call)
> 
> - it's a long running process
> 
> 
> 
> But I will test it as a worker and see if it works.

- If there are multiple comments simultaneously they might create a lot of action.
- I am not sure about this but we would loose a lot of build minutes for this.
- As far as I know, it's 50 outbound requests per request. This could be avoided if we use Cloudflare D1.
- Yeah, it's supposed to be a long process running it again and again doesn't make sense

.github/.ubiquibot-config.yml
gentlementlegen: This file should be removed, otherwise no plugin will run inside of this repo.

0x4007:
GitHub actions are totally free for us. Workers can be costly. 

It is only a problem to do workers if we have to spend money

sshivaditya2019:
> GitHub actions are totally free for us. Workers can be costly. 
> 
> It is only a problem to do workers if we have to spend money

I think there should be a third option apart from workers and actions. Features like this are better suited for a standalone web server.

- That being said total average execution time for this worker is 1ms CPU time and the free plan includes 30 Million ms CPU time.

gentlementlegen:
Tested thsi on my Cloudflare, works fine, thanks for all the bug fixes. In the end switching this plugin from Worker to Action should be as simple as changing a URL to an Action path.

Please remove `.ubiquibot-config.yml` and good to go.

sshivaditya2019:
@gentlementlegen Removed `.ubiquibot-config.yml`. Changed `@ubiquibot` to `ubiquity-os`

gentlementlegen (approved):
Good with me. @0x4007 if you can create the Supabase instance and set the secrets, I do not have permission to create new instances on the company's account, thanks.

0x4007:
I think create and we can transfer later?

Keyrxng:
This is great stuff gg @sshivaditya2019

@gentlementlegen @0x4007 I think it might be a good idea as another issue to improve the db structure so that conversations are grouped together easily. This way when working with the embeddings it's easier to get all of the context specific to a certain task or PR review discussion etc. For instance telegram chat embeddings would be grouped with the issue it relates to in order to give a fuller picture when the embeddings are used.

I guess using the task body is possible but we tend to avoid using arbitrary strings for situations like this

src/types/context.ts
```diff
@@ -3,13 +3,14 @@ import { EmitterWebhookEvent as WebhookEvent, EmitterWebhookEventName as Webhook
...
 import { Logs } from "@ubiquity-dao/ubiquibot-logger";
+import { createAdapters } from "../adapters";
 
 /**
  * Update `manifest.json` with any events you want to support like so:
  *
  * ubiquity:listeners: ["issue_comment.created", ...]
  */
-export type SupportedEventsU = "issue_comment.created";
+export type SupportedEventsU = "issue_comment.created" | "issue_comment.deleted" | "issue_comment.edited";
```
0x4007, line 13: @Keyrxng can't this type be dynamically generated by importing the manifest.json file?

Keyrxng: Unfortunately not without trickery because string arrays type as `string` and inference goes out of the window. 

The easiest way to do it would be to convert them into keys inside the manifest as opposed to a string array this way we could easily extract string literal types from the object.

Issue #8, opened 2024-09-08
Fix Database Schema


0x4007:
This requires changes to the plugin that captures the data, as well as the database itself. 

![2024-09-08 21 54 35](https://github.com/user-attachments/assets/b6885a88-accb-4349-bf39-bb96b890488c)

This must capture issue bodies. 

---

Perhaps I wasn't explicit with my vision for the database schema, but this needs adjustments.

We only need: 

```
id, created_at, modified_at, author_id, plaintext, embedding
```

- The `plaintext` should only save if its a public repo, otherwise we have liabilities on our hands storing company private data. 
- We can make a separate table for when we start capturing messages from Telegram and other platforms. No concerns for now. 
- ID might need to be the GraphQL node id, because, as I understand, the GitHub REST API will not allow us to query the comment with only the numerical ID. We would need to store the organization name, and the repository name, before we can query the comment based on its ID. 
- Dates: the reason for capturing the dates is because I envision that we can make a Telegram chatbot that will allow anybody to query anything about the organization's latest projects in plain English. This is useful for onboarding (new team member has some questions about our infrastructure/products) and investors wanting an update. We need to give less weight to older comments. 
- author id: for onboarding developers especially, it would be useful to target specific contributors' inputs. Mine for final decisions, and others who may be the primary contributors to specific repos, compared to other contributors etc.

sshivaditya2019:
@0x4007 would plaintext include issue body as well or would it just be comment body?

So, right now if there's a private repository then the plaintext would become 'CENSORED'. Is that fine?

0x4007:
Issue body as well and no leave null

0x4007:
Why do you want to associate and/or combine the embedding of the issue body and the comment body?

sshivaditya2019:
> Why do you want to associate and/or combine the embedding of the issue body and the comment body?
> 
> Can you provide some examples of applications that would require this?

Case 1: Issue Body + Comment Body:
- Only one vector index for `Issue Body + Comment Body`
- Might support vector search if we want to search for comments or issues alone.
- Uses: Relevance of the comment to the issue body, Text Based Searching using LIKE

Case 2: Issue Body and Comment Body Separate columns:
- Multiple Vector index required for vector search, as we want to support vector search by issue and comment text
- Multiple Columns (Might not be bad?)
- Uses: TLDR comments, comments/issue based vector search

[...]

PR #9, opened 2024-09-09
Fix: Database Schema

sshivaditya2019:
Resolves [#8](https://github.com/ubiquibot/issue-comment-embeddings/issues/8)

- Updated migrations for the database
- Supports `NULL` comment body for private repository.
- Has timestamps now for creation and modification.
- PKey is Github's global node id. No need for Organization or Repo Name for referencing it.
- Schema has `author_id` now.


src/handlers/add-comments.ts
```diff
@@ -6,20 +6,14 @@ export async function addComments(context: Context) {
...
-  const repo = payload.repository.name;
-  const issueNumber = payload.issue.number;
-  const issueBody = payload.issue.body || "";
-  const owner = payload.repository.owner.login;
-  const body = payload.comment.body;
-
-  // Log the payload
-  logger.info(`Executing addComments:`, { sender, repo, issueNumber, owner });
+  const plaintext = payload.comment.body;
+  const authorId = payload.comment.user?.id || 0;
```
0x4007, line 10:
```suggestion
  const authorId = payload.comment.user?.id || -1;
```

src/types/database.ts
```diff
@@ -30,22 +30,28 @@ export type Database = {
...
-          id: number;
-          issuebody: string | null;
+          id: string;
+          modified_at: string;
+          plaintext: string;
         };
         Insert: {
-          commentbody: string;
+          author_id: string;
+          created_at?: string;
```
0x4007, line 42: Sync with the GitHub metadata of the comment or issue

tests/__mocks__/strings.ts
```diff
@@ -15,4 +15,5 @@ export const STRINGS = {
   CONFIGURABLE_RESPONSE: "Hello, Code Reviewers!",
   INVALID_COMMAND: "/Goodbye",
   COMMENT_DOES_NOT_EXIST: "Comment does not exist",
+  CENSORED: "[CENSORED]",
```
0x4007, line 18:
```suggestion
```

src/adapters/supabase/helpers/comment.ts
```diff
@@ -25,10 +28,11 @@ export class Comment extends SuperSupabase {
       return;
     } else {
       //Create the embedding for this comment
-      const embedding = await this.context.adapters.openai.embedding.createEmbedding(commentBody);
-      const { error } = await this.supabase
-        .from("issue_comments")
-        .insert([{ id: commentId, commentbody: commentBody, issuebody: issueBody, embedding: embedding }]);
+      const embedding = await this.context.adapters.openai.embedding.createEmbedding(plaintext);
+      if (isPrivate) {
+        plaintext = "CENSORED";
```
0x4007, line 33:
```suggestion
        plaintext = null;
```
Should be initialized as null though.

src/adapters/supabase/helpers/comment.ts
```diff
@@ -37,25 +41,28 @@ export class Comment extends SuperSupabase {
...
   }
 
-  async updateComment(commentBody: string, commentId: number) {
+  async updateComment(plaintext: string, commentNodeId: string, isPrivate: boolean) {
     //Create the embedding for this comment
-    const embedding = Array.from(await this.context.adapters.openai.embedding.createEmbedding(commentBody));
-    const { error } = await this.supabase.from("issue_comments").update({ commentbody: commentBody, embedding: embedding }).eq("id", commentId);
+    const embedding = Array.from(await this.context.adapters.openai.embedding.createEmbedding(plaintext));
+    if (isPrivate) {
+      plaintext = "CENSORED";
```
0x4007, line 48:
```suggestion
      plaintext = null;
```

0x4007 (changes requested):

0x4007:
Why did you switch to main from development

sshivaditya2019:
> Why did you switch to main from development that doesn't seem right

Fixed

src/types/database.ts
sshivaditya2019: This is according to the schema mentioned in the issue spec, could you please be more specific with what needs to be synced ?

sshivaditya2019:
@0x4007 Updated the schema, comments use `voyageai` for embeddings now.

src/adapters/openai/helpers/embedding.ts
```diff
@@ -11,15 +11,19 @@ export class Embedding extends SuperOpenAi {
...
-    const response = await this.client.embeddings.create({
-      ...params,
-    });
-    return response.data[0]?.embedding;
+  async createEmbedding(text: string | null): Promise<number[]> {
+    if (text === null) {
+      return new Array(VECTOR_SIZE).fill(0);
+    } else {
+      const params: OpenAI.EmbeddingCreateParams = {
+        model: "text-embedding-3-large",
```
0x4007, line 19: Should this file be deleted or something?

src/adapters/supabase/helpers/comment.ts
```diff
@@ -25,10 +28,18 @@ export class Comment extends SuperSupabase {
       return;
     } else {
       //Create the embedding for this comment
-      const embedding = await this.context.adapters.openai.embedding.createEmbedding(commentBody);
+      const embedding = await this.context.adapters.voyage.embedding.createEmbedding(plaintext);
+      //If embedding is smaller than 3072, pad it with 0s
```
0x4007, line 32: Why does it have to be this exact length?

tests/__mocks__/adapter.ts
```diff
@@ -4,47 +4,56 @@ import { Embedding } from "../../src/adapters/openai/helpers/embedding";
 import { STRINGS } from "./strings";
 
 export interface CommentMock {
-  id: number;
-  commentbody: string;
-  issuebody: string;
+  id: string;
+  plaintext: string | null;
+  author_id: number;
+  commentobject?: Record<string, unknown> | null;
```
0x4007, line 10:
```suggestion
  commentObject?: Record<string, unknown> | null;
```

Our commit hook shouldn't have let this pass because we have strict pascal case and cspell

0x4007 (commented):
Seems generally okay. Let me see how your database looks

src/adapters/openai/helpers/embedding.ts
sshivaditya2019: I’ve retained it in case it can be expanded to support OpenAI's chat APIs. I believe this could be useful, even if it's not currently in use.

src/adapters/supabase/helpers/comment.ts
sshivaditya2019: This is the maximum possible length. Even if we switch providers in the future, we can pad with zeros to accommodate the change. It would be much more challenging to do the opposite.

tests/__mocks__/adapter.ts
sshivaditya2019: I wanted to keep it consistent with the DB schema. This passed the cspell check because I had added it to the ignore list.

src/adapters/supabase/helpers/comment.ts
0x4007: > This is the maximum possible length. 
Seems unnecessary. 

> It would be much more challenging to do the opposite.

Just delete the trailing zeros.

src/adapters/openai/helpers/embedding.ts
0x4007: We can restore from git history if needed. Its not in use so it should be deleted.

tests/__mocks__/adapter.ts
0x4007: > with the DB schema

If you defined it, then its wrong. If its from another database that we can't change, then we can add to the ignore list.

sshivaditya2019:
@0x4007 Could you please check the updated changes ? Have removed `OpenAI` adapter, corrected the syntax and updated the dimension for the `embedding` column ?

0x4007:
You'll need to cherry pick changes (easy to do with a git UI) and deliver to each task separately. The reason why we do this is for audits and review. When there is a problem some time in the future, the developers git blame and review the full conversation and pull for context on debugging. Combining multiple deliverables in a single pull complicates this process and is forbidden.

sshivaditya2019:
> You'll need to cherry pick changes (easy to do with a git UI) and deliver to each task separately. The reason why we do this is for audits and review. When there is a problem some time in the future, the developers git blame and review the full conversation and pull for context on debugging. Combining multiple deliverables in a single pull complicates this process and is forbidden.
> 
> Make sure your CI is passing and make sure to post QA links (show it working on your own org/repos)

I have removed the changes for the other task. CI Should be passing now. 

https://github.com/user-attachments/assets/b79bdf2b-58eb-4618-b9eb-ce2376dcf313

0x4007:
Where's the issue body? You should probably make another table and capture them there. Also might need to associate the comment to the issue with a foreign key

sshivaditya2019:
> Where's the issue body? You should probably make another table and capture them there. Also might need to associate the comment to the issue with a foreign key

Right now issue body is in the payload object, this is jsonb, which can be queried using sql.

image

sshivaditya2019:
So to sum up, you require two tables one for `comments` and one for `issues`. 

The schema for `comments` is `id, plaintext, embedding, payloadobject, type, created_at, modified_at, issue_id`
and the schema for `issues` is `id, plaintext, embedding, payloadobject, type, created_at, modified_at`

Is that right ?

0x4007:
I'm bad at deciding this sort of thing. Let's go with your suggestion for now except the column should be renamed just to payload or context

sshivaditya2019:
https://github.com/user-attachments/assets/ceb63573-4a88-47fb-8c7a-f26bd3adef80

supabase/migrations/20240912225853_issue_comments.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE vectordump
+RENAME TO issue_comments;
+
+CREATE TABLE IF NOT EXISTS issues (
+    id VARCHAR primary key,
+    plaintext text,
+    embedding Vector(1024) not null,
+    payload jsonb,
+    author_id VARCHAR not null,
+    type text not null default 'issue',
```
0x4007, line 10: Why is the default issue on the comments table? Why even add the type column if it's in the comments table?

0x4007:
> @0x4007 Have Added Two Separate Tables as per the schema mentioned in the previous comment. 
> 
> 
> 
> https://github.com/user-attachments/assets/ceb63573-4a88-47fb-8c7a-f26bd3adef80
> 
> 
> 
> 

Seems mostly good but I didn't see all the headers on the first table. Does it have the payload? Just try and normalize the columns as much as you can. 

Otherwise not sure why you added the type column if they are separated by type per table.

supabase/migrations/20240912225853_issue_comments.sql
sshivaditya2019: Its there to ignore, bot comments, and chore issues created. So, while doing vector search we can ignore such issues and comments. I can remove that, but its a feature that could be useful while performing vector search.

sshivaditya2019:
> > @0x4007 Have Added Two Separate Tables as per the schema mentioned in the previous comment.
> > 
> >   
> >     
> >     
> > 
> >     Screen.Recording.2024-09-13.at.1.01.47.AM.mov
> 
> Seems mostly good but I didn't see all the headers on the first table. Does it have the payload? Just try and normalize the columns as much as you can.
> 
> Otherwise not sure why you added the type column if they are separated by type per table.

Yes, it includes a payload. I've retained the type column to distinguish between bot comments and bot issues. However, I can remove it if needed, as implementing that feature might be beyond the scope of this issue.

0x4007:
Yes I think its unnecessary if they are separated by type on different tables.

sshivaditya2019:
Removed the `type` from schema. Payload is stored for both comments and issues.

https://github.com/user-attachments/assets/0a230a7e-69a5-4843-b33b-c03b1d1ce620

0x4007:
Thanks for the thorough QA. You don't need to make a new video on every change! But generally when opening a pull or making major changes a video is useful. 

The last idea I have (sorry for the last second changes) is to have two columns for the text `plaintext` and `markdown`

This is so we can easily do testing in the near future to compare the performance of the plaintext and markdown versions of each comment when reasoning with the LLMs. However the new ChatGPT model `o1` just came out today and is supposed to be very good at reasoning, with built-in chain-of-thought reasoning capabilities. This makes me more optimistic about working with the raw markdown source code, as it provides more context (i.e. blockquotes)

Once that is implemented, you don't need to make a QA video. Just let me know and we can merge. Do that for both tables please.

sshivaditya2019:
@0x4007 I have added `markdown` and `plaintext` column.  `o1` has a great reasoning capabilities, I think ChatGPT-Plus members have access to it already, probably will take a lot of time to be GA and be available on API.

src/adapters/utils/markdown-to-plaintext.ts
0x4007, line 1: This seems very error prone. Why dont you use some virtual DOM (jsdom, mdast) and then just query `element.textContent` ?

src/adapters/voyage/helpers/embedding.ts
```diff
@@ -0,0 +1,25 @@
...
+    this.context = context;
+  }
+
+  async createEmbedding(text: string | null): Promise<number[]> {
+    if (text === null) {
+      return new Array(VECTOR_SIZE).fill(0);
+    } else {
+      const response = await this.client.embed({
+        input: text,
+        model: "voyage-2",
```
0x4007, line 20:
```suggestion
        model: "voyage-large-2-instruct",
```

Should be using the best model. 

> Top of [MTEB leaderboard](https://huggingface.co/spaces/mteb/leaderboard). Instruction-tuned general-purpose embedding model optimized for clustering, classification, and retrieval. For retrieval, please use input_type parameter to specify whether the text is a query or document. For classification and clustering, please use the instructions [here](https://github.com/voyage-ai/voyage-large-2-instruct). See [blog post](https://blog.voyageai.com/2024/05/05/voyage-large-2-instruct-instruction-tuned-and-rank-1-on-mteb/) for details.

[Source](https://docs.voyageai.com/docs/embeddings)

src/handlers/add-issue.ts
```diff
@@ -0,0 +1,29 @@
+import { Context } from "../types";
+import { IssuePayload } from "../types/payload";
+
+export async function addIssue(context: Context) {
+  const {
+    logger,
+    adapters: { supabase },
+  } = context;
+  const { payload } = context as { payload: IssuePayload };
+  const markdown = payload.issue.body + " " + payload.issue.title || "";
```
0x4007, line 10:
```suggestion
  const markdown = payload.issue.body + " " + payload.issue.title || null;
```

Empty string seems like an impossibility. You can't post empty comments so this actually should probably throw an error.

src/handlers/add-issue.ts
```diff
@@ -0,0 +1,29 @@
...
+      logger.error(`Error creating issue:`, { error: error, stack: error.stack });
+      throw error;
+    } else {
+      logger.error(`Error creating issue:`, { err: error, error: new Error() });
+      throw error;
+    }
+  }
+
+  logger.ok(`Successfully created issue!`);
+  logger.verbose(`Exiting addIssue`);
```
0x4007, line 28:
```suggestion
  logger.debug(`Exiting addIssue`);
```

I think we have a debug method which might be more appropriate for these sort of messages in the future for you.

0x4007:
> I think ChatGPT-Plus members have access to it already, probably will take a lot of time to be GA and be available on API.

I think that only tier5 subscribers can use right now. I believe that we are tier4.

src/adapters/utils/markdown-to-plaintext.ts
sshivaditya2019: I don't think its safe to give codeblock to jsdom. This works fine, I have tested with several examples now. Basically removes the delimiters and extracts the text inside them.

sshivaditya2019: Replace it with `Markdown-it`. This library is being maintained actively and has sanitization inbuilt.

src/adapters/voyage/helpers/embedding.ts
```diff
@@ -0,0 +1,25 @@
...
+  protected context: Context;
+
+  constructor(client: VoyageAIClient, context: Context) {
+    super(client, context);
+    this.context = context;
+  }
+
+  async createEmbedding(text: string | null): Promise<number[]> {
+    if (text === null) {
+      return new Array(VECTOR_SIZE).fill(0);
```
0x4007, line 16: Still confused about this fill stuff

sshivaditya2019: If a text is empty, instead of calling embedding api on it I am using fill to create zeros array of 1024-dim. We have to do this as embedding has to be a not null value for vector search to work. And zero vectors do not cause any change to vector search result as well.

src/adapters/supabase/helpers/issues.ts
```diff
@@ -0,0 +1,88 @@
...
+      .textSearch("embedding", embedding.join(","))
+      .order("embedding", { foreignTable: "issues", ascending: false })
+      .lte("embedding", threshold);
+    if (error) {
+      this.context.logger.error("Error finding similar issues", error);
+      return [];
+    }
+    return data;
+  }
+}
```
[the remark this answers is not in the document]
sshivaditya2019: This can be used for issue deduplication and stuff. I think this should be here m.

0x4007 (approved):

src/adapters/supabase/helpers/issues.ts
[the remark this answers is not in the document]
0x4007: Disagree but no need to slow down this pull further.

PR #11, opened 2024-09-13
feat: Issue Deduplication

### LINKED ISSUE #6: Issue Dedupe
On `issue.created` and `issue.edited` we should check if there are any open issues that are x% similar within the repository. There should be two configurable percentages: 
1. A `warning` threshold, perhaps 75% similar as a default 
2. "Match" threshold, 95% similar default. 

- If a "match" is found, the bot should "close as unplanned" 
- If a "warning" level issue has been found, post a comment with all the similar open issues. Target and edit this comment if it currently exists. 

It isn't entirely clear to me if this should be a separate plug-in, but this does require coupling to the same vector embeddings database. Another idea is to pass in the authentication details in each vector embeddings related plugin's config.

sshivaditya2019:
Resolves #6


src/handlers/issue-deduplication.ts
```diff
@@ -0,0 +1,136 @@
...
+    payload,
+    adapters: { supabase },
+  } = context;
+
+  const issue = payload.issue;
+
+  //First Check if an issue with more than MATCH_THRESHOLD similarity exists (Very Similar)
+  const similarIssue = await supabase.issue.findSimilarIssues(issue.body + issue.title, MATCH_THRESHOLD, issue.node_id);
+  if (similarIssue && similarIssue?.length > 0) {
+    logger.info(`Similar issue which matches more than ${MATCH_THRESHOLD} already exists`);
```
0x4007, line 30: Needs to return the results of issues

sshivaditya2019: @0x4007 This function handles only the deduplication process, if there's a match it closes the issue, if there's a warning condition, it would create the comment. I am not sure why the similarIssue would be needed in other functions like `add-comment`, etc.

sshivaditya2019:
@0x4007  I have tried to make a few examples, let me know if have more or want to test more.

https://github.com/user-attachments/assets/ac8c81dd-f70e-4a77-a86a-3cb752bad000

src/handlers/issue-deduplication.ts
0x4007: It needs to always let the user know which it thinks is similar. It can be wrong

0x4007:
Adding labels is out of scope. Don't do that. Close it as unplanned, don't add any labels.

0x4007 (changes requested):
- Adding labels is out of scope. Don't do that. Close it as unplanned, don't add any labels. 
- Add a match percentage as well when any are listed. 
- How did you generate the test cases and determine their percentage similarity?

src/handlers/issue-deduplication.ts
```diff
@@ -0,0 +1,138 @@
+import { IssueSimilaritySearchResult } from "../adapters/supabase/helpers/issues";
+import { Context } from "../types";
+const MATCH_THRESHOLD = 0.95;
+const WARNING_THRESHOLD = 0.5;
```
0x4007, line 4: Why did you do 50%?

src/handlers/issue-deduplication.ts
```diff
@@ -0,0 +1,138 @@
...
+  const similarIssue = await supabase.issue.findSimilarIssues(issue.body + issue.title, MATCH_THRESHOLD, issue.node_id);
+  if (similarIssue && similarIssue?.length > 0) {
+    logger.info(`Similar issue which matches more than ${MATCH_THRESHOLD} already exists`);
+    //Close the issue as "unplanned"
+    await context.octokit.issues.update({
+      owner: payload.repository.owner.login,
+      repo: payload.repository.name,
+      issue_number: issue.number,
+      state: "closed",
+      labels: ["unplanned"],
```
0x4007, line 37: No labels

src/handlers/issue-deduplication.ts
```diff
@@ -0,0 +1,138 @@
...
+                    }`,
+          {
+            issueNodeId: issue.issue_id,
+          }
+        );
+        return issueUrl;
+      })
+    );
+
+    // Reopen the issue
```
0x4007, line 68: What is going on here? Reopening issues is out of scope.

sshivaditya2019:
> * Adding labels is out of scope. Don't do that. Close it as unplanned, don't add any labels.
> * Add a match percentage as well when any are listed.

Added They will display the cosine similarity in percentage after each issue in the list.

> * How did you generate the test cases and determine their percentage similarity?

I manually calculated and created test cases using embeddings and found their cosine similarity values.

src/handlers/issue-deduplication.ts
sshivaditya2019: A cosine similarity of 0.75 appears quite close for identifying similar issues. I tested this with a few examples and noticed some potential errors with the samples. Typically, for similar issues, the similarity was either above 75% and aligned with 95% category or around 60%. Therefore, I experimented with a 50% threshold, which seemed to work well.

src/handlers/issue-deduplication.ts
sshivaditya2019: Removed.

src/handlers/issue-deduplication.ts
sshivaditya2019: Removed Labels. Labels will not be added on issue close

src/handlers/issue-deduplication.ts
sshivaditya2019: So, for both `Match` and `Warning` threshold, there has to be comment with similar issues. Is that right ?

0x4007:
Can you link your issue where you tested so we can see the results?

sshivaditya2019:
> Can you link your issue where you tested so we can see the results?

95%:
- [First Comment](https://github.com/sshivaditya2019/test-public/issues/81)
- [Second Comment](https://github.com/sshivaditya2019/test-public/issues/80)

50%:
- [First Comment](https://github.com/sshivaditya2019/test-public/issues/78)
- [Second Comment](https://github.com/sshivaditya2019/test-public/issues/79

0x4007:
Okay it seems like you aren't following the spec again. 

Needs to list the similar results on every scenario. 

Do 75% as a default.

sshivaditya2019:
> Okay it seems like you aren't following the spec again.
> 
> Needs to list the similar results on every scenario.

Fixed that, it now returns the similar issue in both `MATCH` and `WARNING` case.

> 
> Do 75% and 95% as a default.

Warning Threshold is 75% now.

95%:
- [First Comment](https://github.com/sshivaditya2019/test-public/issues/88)
- [Second Comment](https://github.com/sshivaditya2019/test-public/issues/89)

75%:
- [First Comment](https://github.com/sshivaditya2019/test-public/issues/90)
- [Second Comment](https://github.com/sshivaditya2019/test-public/issues/91)

0x4007:
> Fixed that, it now returns the similar issue in both MATCH 

Doesn't look like it

sshivaditya2019:
> * First Comment

That's the first issue of that type, so its expected to not have similar issues. Two, should be the first time a similar issue, is found with similarity more than 95%.

src/handlers/issue-deduplication.ts
```diff
@@ -0,0 +1,114 @@
+import { IssueSimilaritySearchResult } from "../adapters/supabase/helpers/issues";
+import { Context } from "../types";
+import { IssuePayload } from "../types/payload";
+
+const MATCH_THRESHOLD = 0.95;
+const WARNING_THRESHOLD = 0.75;
```
0x4007, line 6: These should be configurable values. Can you see how configuration is implemented in the other plugins?

0x4007 (commented):
Cool just needs configuration and I can merge.

src/handlers/issue-deduplication.ts
sshivaditya2019: Added, Now it fetches the values from the `context.config`

0x4007 (approved):
I'm assuming it all works. Code looks good.

PR #16, opened 2024-09-18
Refactor db

### LINKED ISSUE #13: Store issue if not stored already
1. If the issue does not already exist in the DB then store it to allow comments to be stored.
2. Correct the error handling; when an error is thrown an `ok` log shouldn't be shown if it's not `ok`.
3. Post a comment on the issue stating that the plugin has failed?

image

### LINKED ISSUE #15: Handle HTML input
Input should be properly sanitized and then stored according to the database schema. It appears that markdown is handled correctly but HTML is not and it should be as it's a supported GitHub comment format.

https://github.com/ubq-testing/generate-vector-embeddings/issues/5

image

Keyrxng:
Resolves #13
Resolves #15 

- combined into a single table for embedding storage
- combine adapters into one single `embeddings` entity
- Added mentlegen's proxy with my types to get rid of lots the need for lots of typeguards for each webhook callback
- no longer storing `body > title`, so to make up for that I simSearch using the title and body separately
- refactored tests to use the `mswjs` DB and handlers like other plugins tend to do (mocking modules kills me for some reason, but the db and handlers I understand 😂) 

QA:

###### ignore the row id 8 with the "::", I added `body > title` logic and removed it immediately.
I have another pull with the `setup_instructions` logic included, left out as it was not in scope for this PR.

image


gentlementlegen (commented):
I am fine with the code changes themselves, but I am not sure what to test here. @Keyrxng could you give me a use case?

Keyrxng:
> I am fine with the code changes themselves, but I am not sure what to test here. @Keyrxng could you give me a use case?

I'm unsure what you mean a use case? Like to practically work with the embeddings in unison with an LLM?

If I was testing this PR I'd ensure that the comments and specs are being stored properly. You could create embeddings for a spec or a readme then create a query embedding and see far fruitful it is with that alone. Although chatbot/embedding effectiveness I'd think is sort of out of scope for this. The intention of the PR was to make it easier to work with the embeddings from more than one angle.

The class system should make it easier for us to refine our embedding scope before we compare, that way we improve it's accuracy because we can remove entire "sections" of data and have it only consider what we tell it to. The infra isn't in place yet of course, but the class system for e.g we install a chatbot on ubq.fi, easy to do and good value for money, it should be scoped to high level org stuff and guard railed.

Moreover, the class system can be used as a weighted tier system for prioritizing certain sections over others under certain conditions.

https://github.com/ubiquibot/generate-vector-embeddings/pull/18 introduces logic that'll gather readmes which would give us something that we could effectively test in terms of having a well-rounded sort of chatbot with actionable milestones for confirming it effectiveness.


I'm not sure I answered your question or not please let me know

gentlementlegen:
@Keyrxng thank  you for the clarification. To ask simply, I wondered what I should be looking for when testing, the comments that contain HTML should be stripped from their tags and stored as raw text in the db is that correct?

0x4007:
I think we should store with full source code and then stripped just text and compare the results. 

But eventually we should settle on one way and I'm pretty sure it's gonna end up being with the full markup

src/adapters/utils/markdown-to-plaintext.ts
```diff
@@ -6,13 +6,16 @@ import plainTextPlugin from "markdown-it-plain-text";
  * @param markdown
  * @returns
  */
-export function markdownToPlainText(markdown: string | null): string | null {
+export function markdownToPlainText(markdown?: string | null): string {
   if (!markdown) {
-    return markdown;
+    return "";
```
0x4007, line 11: Is this right?

src/handlers/create-comment-embedding.ts
```diff
@@ -0,0 +1,14 @@
...
+export async function createCommentEmbedding(context: Context<"issue_comment.created">): Promise<CallbackResult> {
+  const {
+    logger,
+    adapters: { supabase },
+  } = context;
+
+  const uploaded = await supabase.embeddings.createConversationEmbeddings(context.payload.comment.node_id, context.payload, "comment");
+  logger.ok(`Successfully created comment!`, { ...uploaded, embedding: "removed for brevity" });
+
+  return { status: 200, reason: "success" };
```
0x4007, line 13: What's the purpose of reason property 

200 is self explanatory

Keyrxng:
> I think we should store with full source code and then stripped just text

I agree, I think we ought to refer to others in the space and see how they are doing it.

- [Cursor](https://forum.cursor.com/t/codebase-indexing/36)
- [Continue](https://docs.continue.dev/customize/deep-dives/codebase)
- [QDrant](https://qdrant.tech/documentation/tutorials/code-search/)
>Data preparation
Chunking the application sources into smaller parts is a non-trivial task. In general, functions, class methods, structs, enums, and all the other language-specific constructs are good candidates for chunks. They are big enough to contain some meaningful information, but small enough to be processed by embedding models with a limited context window. You can also use docstrings, comments, and other metadata can be used to enrich the chunks with additional information.

> Parsing the codebase
While our example uses Rust, you can use our approach with any other language. You can parse code with a [Language Server Protocol](https://microsoft.github.io/language-server-protocol/) (LSP) compatible tool. You can use an LSP to build a graph of the codebase, and then extract chunks. We did our work with the [rust-analyzer](https://rust-analyzer.github.io/). We exported the parsed codebase into the [LSIF](https://microsoft.github.io/language-server-protocol/specifications/lsif/0.4.0/specification/) format, a standard for code intelligence data. Next, we used the LSIF data to navigate the codebase and extract the chunks. For details, see our [code search demo](https://github.com/qdrant/demo-code-search).

Everyone does something similar but different although they all chunk their codebase, split it into semantically effective sections. So we need to determine how we are going to prep our codebase(s) for chunking, if we'll do it all manually or involve some other recommended tooling or whatever.

src/adapters/utils/markdown-to-plaintext.ts
Keyrxng: I guess empty strings are avoided so no it's not

src/handlers/create-comment-embedding.ts
gentlementlegen: I think going forward it would make more sense to use `status` and eventually `message` if we wanna convey data.

0x4007: We should copy other popular frameworks and use `statusCode`

Keyrxng:
image

PR #19, opened 2024-09-23
feat: schema validation

gentlementlegen:
Resolves https://github.com/ubiquity/ubiquibot-kernel/issues/80


tests/main.test.ts
```diff
@@ -37,7 +37,7 @@ describe("Plugin tests", () => {
 
   it("Should serve the manifest file", async () => {
     const worker = (await import("../src/worker")).default;
-    const response = await worker.fetch(new Request("http://localhost/manifest.json"), {
+    const response = await worker.fetch(new Request("http://localhost/manifest"), {
```
0x4007, line 40: I don't like this removal of the file name extension.

gentlementlegen: It's very strange to hit endpoints with extensions except to download files usually.

PR #20, opened 2024-09-25
Task matching

### LINKED ISSUE #7: Task Matchmaking
When a new task is posted to our network, UbiquityOS should recommend qualified contributors. 

- The system must understand the "closed as complete" project specifications. 
- this should run when 
   - an issue is priced, so I believe on `issues.label_added`
   - issue created `issues.created`
   - task specification has been edited `issues.edited `
- if the new task is 75% similar or greater (configurable) it should post a comment with a list of those issues in order of percentage match. This comment SHOULD NOT tag the users. Instead this is intended for a human to check their work, and tag them in as needed. 
- Be sure to add the `www.` prefix on the github href so that it does not register as a "link back" on the GitHub UI over on their issue.  

For example it should look something like:

```
### 0x4007
- ubiquity/ubiquity-dollar#123 87% match. 
### rndquu
- ubiquity/ubiquity-dollar#250 77% match. 
```

- if the search results comment exists, target and edit it instead of reposting. This is relevant when editing the specification. 

Eventually this may be configurable to be network wide (global) organization wide, or repository wide. But for this first version, we will do network wide.

---

Another configuration of this idea could be to recommend the contributor in the organization with the highest match, even if it doesn't hit the threshold. If within the organization, perhaps tagging is appropriate. Perhaps it makes sense to handle this in a separate task.

sshivaditya2019:
Resolves #7 

-  Adds a `issue` matching option. 
-  Works on `issue.labeled`, `issues.created` and `issues.edited`.
-  Updates the payload object, when someone is assigned a particular issue.


sshivaditya2019:
Example issue:

[New Issue](https://github.com/sshivaditya2019/test-public/issues/90)
[Old Issue](https://github.com/sshivaditya2019/test-public/issues/90)

Should it search for only closed issues ? Or Should i check in progress issues as well ? 

So, for the `closed as complete",  we would have to listen to `issues.closed` webhook, or should we use the linked PR to check if it was `closed as complete`?

0x4007:
> Updates the payload object, when someone is assigned a particular issue.

I don't understand what you mean by this.

> Should it search for only closed issues ? Or Should i check in progress issues as well ?

Only closed as completed issues. This proves that they were able to accomplish the task. 

> So, for the closed as complete, we would have to listen to issues.closed webhook, or should we use the linked PR to check if it was closed as complete?

I don't understand. No webhooks are applicable here. You use GraphQL to search all the "closed as complete" issues that are assigned to the contributor.

sshivaditya2019:
> > Updates the payload object, when someone is assigned a particular issue.
> 
> 
> 
> I don't understand what you mean by this.
> 
> 

If some is assigned an issue that particular issue's payload is updated. This could be rewritten with graphql instead.

> 
> > Should it search for only closed issues ? Or Should i check in progress issues as well ?
> 
> 
> 
> Only closed as completed issues. This proves that they were able to accomplish the task. 
> 
> 
> 
> > So, for the closed as complete, we would have to listen to issues.closed webhook, or should we use the linked PR to check if it was closed as complete?
> 
> 
> 
> I don't understand. No webhooks are applicable here. You use GraphQL to search all the "closed as complete" issues that are assigned to the contributor. 
> 
> 

Shouldn't it be a query with similar issues and who completed them. Instead of querying with all users.

So, when issues are completed their payload object will still show as open. To prevent this we could listen for webhooks.

0x4007:
I'm sorry but I have little context on the "payload object" I haven't worked on any plugins officially. Feel free to ask me clarifications on the spec.

Perhaps @gentlementlegen might know what you're talking about

sshivaditya2019:
Fixed, Now, its using graphQL for query, listens only for `issues.labled`, `issues.opened` and `issues.edited`. For QA

[New Issue](https://github.com/sshivaditya2019/test-public/issues/90)
[Old Issue](https://github.com/sshivaditya2019/test-public/issues/90)

0x4007:
> Fixed, Now, its using graphQL for query, listens only for `issues.labled`, `issues.opened` and `issues.edited`. For QA
> 
> [New Issue](https://github.com/sshivaditya2019/test-public/issues/90) [Old Issue](https://github.com/sshivaditya2019/test-public/issues/90)

QA looking good. Please check my comment under "old issue" regarding formatting! Once thats in I would like to merge.

sshivaditya2019:
Updated the comment structure, the comments should start with `>[!NOTE]` , followed by the ">The following contributors may be suitable for this task:", and then  match percentage

src/handlers/issue-matching.ts
```diff
@@ -0,0 +1,150 @@
...
+        login: string;
+        url: string;
+      }>;
+    };
+  };
+  similarity: number;
+}
+
+const commentBuilder = (matchResultArray: Map<string, Array<string>>): string => {
+  const commentLines: string[] = [">[!NOTE]", ">The following contributors may be suitable for this task:"];
```
0x4007, line 28:
```suggestion
  const commentLines: string[] = ["> [!NOTE]", "> The following contributors may be suitable for this task:"];
```

src/handlers/issue-matching.ts
```diff
@@ -0,0 +1,150 @@
...
+  matchResultArray.forEach((issues, assignee) => {
+    commentLines.push(`>### [${assignee}](https://www.github.com/${assignee})`);
```
0x4007, line 30:
```suggestion
    commentLines.push(`> ### [${assignee}](https://www.github.com/${assignee})`);
```

0x4007 (commented):
Did you test this? The syntax doesn't look correct. Pretty sure all quotes need the space to avoid problems

src/handlers/issue-matching.ts
sshivaditya2019: Adding the space, somehow breaks it. I tried without space, it does not render the note.

sshivaditya2019:
> Did you test this? The syntax doesn't look correct. Pretty sure all quotes need the space to avoid problems

Its working, you can check the updated comment in the ([Old Issue](https://github.com/sshivaditya2019/test-public/issues/90))

src/handlers/issue-matching.ts
sshivaditya2019: It works well both with and without the space, based on my tests.
>[!NOTE]
>Hello
>### Hello
> ### Hello

0x4007:
@gentlementlegen can you install this plugin

gentlementlegen:
@0x4007 Is this plugin supposed to be a worker or an action? According to the readme it should be an action but not deployment script has been added here.

sshivaditya2019:
> run crashing

It should be deployed as an worker. I'll update the readme.

gentlementlegen:
Added deployment scripts, last run [here](https://github.com/ubiquity-os-marketplace/generate-vector-embeddings/actions/runs/11093488654), will change to a worker in the configuration.

gentlementlegen:
Another question, I am confused by [these logs](https://github.com/ubiquity-os-marketplace/generate-vector-embeddings/actions/runs/11093507537/job/30819872800#step:5:9). Does it mean it is running properly?

gentlementlegen:
@0x4007 I do not see a database set up on Supabase for this. Could you create an instance? I do not have the permissions to do so.

0x4007:
You can make a separate database. We can clean house another week

gentlementlegen:
You mean that I should use my own credentials / API keys for this db?

gentlementlegen:
I upgraded with my own credentials. Heads up that now it seems you're limited to two projects on Supabase.

0x4007:
I think just make one ubiquity-os project on your Supabase and then make new tables with the names of the plugins I guess.

gentlementlegen:
This would need a centralized migration file to keep the db up to date. Projects with dbs already have their own migration files so all projects should be merged.

0x4007:
I prefer figuring out git based storage as a "final" storage solution. Let's make do in a janky way with Supabase for the short term.

gentlementlegen:
If there is an efficient way to store, access, sort and join millions of items based on a git system, this can be replaced. Vector embeddings should have a consequent size quickly. I think in the meantime each project should use their own db when possible.

---

The project seems properly set up although it always fails to create entries, see this run. @sshivaditya2019 Am I doing something wrong?

sshivaditya2019:
> If there is an efficient way to store, access, sort and join millions of items based on a git system, this can be replaced. Vector embeddings should have a consequent size quickly. I think in the meantime each project should use their own db when possible.
> 
> The project seems properly set up although it always fails to create entries, see this run. @sshivaditya2019 Am I doing something wrong?

The `issue_id` serves as the foreign key, meaning the corresponding issue should exist in the database before the comment is created.

gentlementlegen:
So how is an issue created?

sshivaditya2019:
> So how is an issue created?

Currently, the expected process is as follows:

An issue is created, followed by comments that can be created, edited, or deleted. When an issue is deleted, all associated comments are also removed.

gentlementlegen:
Okay it does work when an issue is created, I do get new entries in the `issue_comments` table. Is `plaintext` always supposed to be `NULL`?
image
image

sshivaditya2019:
It relates to its execution in GitHub Actions. `plaintext` is not supposed to be null, and neither is `payload`.

gentlementlegen:
@sshivaditya2019 It is currently running within `Workers` since I know it runs properly now, so I don't think that is the reason.

sshivaditya2019:
I am trying to deploy it on worker right now with the workers script. In the meantime could you please try `https://ubiquity-os-comment-vector-embeddings.sshivaditya.workers.dev` this.  I can check if issues are begin created here or not.

gentlementlegen:
@sshivaditya2019 It is a private repository indeed, I forgot about the redacting of the comments, it makes sense now. I will try with a public repo.

0x4007:
Maybe we shouldn't redact comments in the near future. And instead have a switch for it later when/if partners ever ask. I feel like most people won't care and it will give us more data to work with for research purposes in the near future.

0x4007:
@sshivaditya2019 do you think we should make a new task to create a script to populate our database with the issues? We have a ton of old data we can benefit from.

gentlementlegen:
> Maybe we shouldn't redact comments in the near future. And instead have a switch for it later when/if partners ever ask. I feel like most people won't care and it will give us more data to work with for research purposes in the near future.

I guess it would be fine to have it as an option in the configuration, because as long as not everybody can access the DB the content of private repositories cannot be read by outsiders.

PR #22, opened 2024-09-30
fix: enable RLS for tables

gentlementlegen:
Added RLS security for tables to avoid having the data readable / writable by anyone.


gentlementlegen:
@sshivaditya2019 rfc

0x4007 (commented):
I don't think this really matters much if the auth token is hidden.

gentlementlegen:
That's true, it is more to avoid Supabase warnings and updated the `README` with it as well.

README.md
```diff
@@ -12,7 +12,7 @@ To set up the `.dev.vars` file, you will need to provide the following variables
 ## Usage
 - Add the following to your `.ubiquibot-config.yml` file with the appropriate URL:
 ```yaml
-- plugin: ubiquity-os-marketplace/generate-vector-embeddings
+- plugin: https://ubiquity-os-comment-vector-embeddings-main.ubiquity.workers.dev
```
0x4007, line 15: I just realized maybe for the workers URLs it makes sense to do this:

```suggestion
- plugin: https://comment-vector-embeddings.ubiquity-os.workers.dev
```

We now have a dedicated GitHub organization to host the plugins (@ubiquity-os-marketplace) so we may no longer need the `ubiquity-os-` prefix in the plugin name?

0x4007 (approved):

README.md
gentlementlegen: Yes you can remove the `ubiquity-os` prefix. We have to figure out how to handle the domain url, let's keep discussing this on the other pull-request you opened.

Issue #23, opened 2024-09-30
Redact switch


0x4007:
Config to toggle redacts. 

Used for private repos only. 

redactPrivateRepoComments: boolean 

Default we capture everything. 

> Maybe we shouldn't redact comments in the near future. And instead have a switch for it later when/if partners ever ask. I feel like most people won't care and it will give us more data to work with for research purposes in the near future.

I guess it would be fine to have it as an option in the configuration, because as long as not everybody can access the DB the content of private repositories cannot be read by outsiders.

_Originally posted by @gentlementlegen in https://github.com/ubiquity-os-marketplace/generate-vector-embeddings/pull/20#issuecomment-2382245540_

0x4007:
```
> ```diff
> ! You must be a core team member, or an administrator to start this task
> ```
> 
> <!-- UbiquityOS - checkRequirements - CLOUDFLARE_VERSION_METADATA - @koya0 - https://dash.cloudflare.com/<missing-cloudflare-account-id>/workers/services/view/ubiquity-os-command-start-stop-development/production/observability/logs?granularity=0&time=%7B%22type%22%3A%22absolute%22%2C%22to%22%3A1736803182734%2C%22from%22%3A1736803062734%7D
> {
>   "caller": "checkRequirements"
> }
> -->
> 
```

@gentlementlegen Metadata is missing some values. 

I also think we can shorten the message to "You must be a core team member to start this task"

`missing-cloudflare-account-id`

0x4007:
If they are an admin it is implied that they are a core team member. You can make a dedicated admin only one as well then.

gentlementlegen:
Admin is a core team member, but a core team member is not an admin. You can also have outside contributors that are admin of certain repositories.

The text is dynamic and will display:
- "administrator"
- "core team member"
- "contributor"

and combine these according to the settings. Example: "You must be an administrator to start this task" if admin only, or "You must be a core team member, or an administrator to start this task" such as here when contributors + admins are allowed

0x4007:
I understand. Still, the only two messages should be:

"You must be an admin to start this task"

Or 

"You must be a core team member to start this task"

0x4007:
Yes admins always should have full rights.

0x4007:
I think that's not useful either way because self assignment with the UI is possible for collaborators. 

Also permissions should be like a pyramid. 

Admin at top, with all permissions on the below roles. 

[...]

PR #24, opened 2024-10-01
fix: added tests and redactPrivateRepoComments

sshivaditya2019:
Resolves #23 

- Adds tests for `Issue-Deduplication`
- Adds tests for `Task matching`
- Adds redactPrivateRepoComments in the configuration

### DIFF
--- a/tests/__sample__/match_threshold_95_1.json
+++ b/tests/__sample__/match_threshold_95_1.json
@@ -1,0 +1,4 @@
+{
+    "title": "Bug: NullPointerException in Python API Handler",
+    "issue_body": "Title: Bug: NullPointerException in Python API Handler\nDescription: A NullPointerException occurs in the get_user function of the Python API handler when a null value is passed for user_id. Implement a null check to handle this scenario.\nSteps to Reproduce: Pass a null user_id to get_user. Observe the exception in the API logs.\nExpected Behavior: Null values should be handled without causing an exception.\nCurrent Behavior: The function throws a NullPointerException.\nAttachments: Stack trace log."
+}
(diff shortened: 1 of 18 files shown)


Keyrxng:
> @Keyrxng I’ve added some tests for Deduplication and Task Matching that you can pull into your PR. The tests seem to be working well, and the CI is passing with them.

That's awesome dude taking initiative like that, love to see it. Is this ready for review I guess?

tests/__sample__/match_threshold_95_1.json
0x4007, line 1: I think it makes sense to test with our real specifications.

Keyrxng: > Perhaps there is a way to "train" the system that all of our issues that exist open now are unique?

https://github.com/ubiquity-os/plugins-wishlist/issues/50#issuecomment-2388377812

I believe using this system would be effective. Docs would contain the right context anchors uniformly, then when we fine-tune/train our own model with these docs we train it around these context anchors specifically although they'll be pretty effective without it imo.

If we only have docs for closed tasks then this already effectively treats open tasks as unique and a spec contains the why and typically the how also, which our docs contain for each task, we'll have improved performance in my opinion.

0x4007: Unique means that the plugin should not flag them as redundant. 

This is a tangible test with a boolean result that we can use to tweak our approach.

0x4007 (approved):
This pull actually looks fine whats the status?

0x4007:
@sshivaditya @sshivaditya2019

0x4007 (approved):
In the future, you should separate into two different pulls, one adding tests and one solving the problem because larger pulls obviously take a lot more effort to review.

0x4007:
@sshivaditya2019 there are merge conflicts now so after you fix those, you can merge this in

Keyrxng:
weird

PR #26, opened 2024-10-01
fix: Issue dedup relevance

### LINKED ISSUE #25: Similar Issues Adjustments
> This issue seems to be similar to the following issue(s):
> 
> * [Do not break the loop when a plugin fails to run](https://github.com/ubiquity-os/ubiquity-os-kernel/issues/120) Similarity: 76.84

This ran extremely fast. 

@sshivaditya2019: 

1. can you make sure that this includes the `www.` prefix so it does not "reference" or "link back" to the similar issue? I'm also curious why it thinks its a similar issue.  
2. Also I think it should probably only check within the same repo, instead of the entire organization. 
3. Rounding the percentage seems useful. It doesn't matter to us below a fraction.
4. Style adjustments:

---

> [!NOTE]
> #### Similar Issues:
>
> - `77%` [Do not break the loop when a plugin fails to run](https://www.github.com/ubiquity-os/ubiquity-os-kernel/issues/120) 

---

_Originally posted by @0x4007 in https://github.com/ubiquity-os/plugins-wishlist/issues/52#issuecomment-2386585263_

sshivaditya2019:
Resolves #25 

- Developed a new similarity search function, uses l2 distance  (Euclidean Distance) now.


src/handlers/issue-deduplication.ts
```diff
@@ -69,29 +86,39 @@ async function handleSimilarIssuesComment(context: Context, payload: IssuePayloa
...
   );
 
-  const commentBody = issueList.map((issue) => `- [${issue.node.title}](${issue.node.url}) Similarity: ${issue.similarity}`).join("\n");
-  const body = `This issue seems to be similar to the following issue(s):\n\n${commentBody}`;
+  const commentBody = issueList
+    .filter((issue) =>
+      matchRepoOrgToSimilarIssueRepoOrg(payload.repository.owner.login, issue.node.repository.owner.login, payload.repository.name, issue.node.repository.name)
+    )
+    .map((issue) => {
+      const modifiedUrl = issue.node.url.replace("github.com", "www.github.com");
```
0x4007, line 110:
```suggestion
      const modifiedUrl = issue.node.url.replace("https://github.com", "https://www.github.com");
```

Seems more precise.

0x4007:
4o:

> > I believe that to compute the Dice coefficient, we need to generate bit vectors and perform operations on them.
> 
> Right, I see the challenge with Dice on our continuous data. Since we’ve already tried **Cosine Similarity** and are now testing **Euclidean Distance**, it’s good to explore different approaches.
> 
> Cosine focuses on vector direction, whereas Euclidean captures the magnitude differences. Both have their merits depending on the nature of the vectors (angle vs distance). If **Cosine** wasn’t optimal in previous attempts, I’m curious about how Euclidean compares in performance on this data.
> 
> That said, we might also want to experiment with **Manhattan Distance** or even **Minkowski Distance** as a middle ground between Euclidean and Cosine. They could provide further insight into how distance measures impact our similarity task.

0x4007:
> It would not edit inside the issue specification itself without creating a new comment: 

What does that mean? I can edit anybody's specification with collaborator permissions. We just need to ensure the app has sufficient permissions I'm sure.

src/handlers/issue-deduplication.ts
```diff
@@ -69,42 +86,46 @@ async function handleSimilarIssuesComment(context: Context, payload: IssuePayloa
...
+      matchRepoOrgToSimilarIssueRepoOrg(payload.repository.owner.login, issue.node.repository.owner.login, payload.repository.name, issue.node.repository.name)
+    )
+    .map((issue, index) => {
+      const modifiedUrl = issue.node.url.replace("https://github.com", "https://www.github.com");
+      finalIdx += 1;
+      return `[^0${index + 1}^]: [${issue.node.title}](${modifiedUrl}) ${issue.similarity}%`;
+    })
+    .join("\n");
+  const footnoteLinks = [...Array(finalIdx).keys()].map((i) => `[^0${i + 1}^]`).join("");
+  const body = "\n###### Similar " + footnoteLinks + ":\n\n" + commentBody;
```
0x4007, line 117:
```suggestion
  const body = "\n###### Similar " + footnoteLinks + "\n\n" + commentBody;
```

src/handlers/issue-deduplication.ts
```diff
@@ -69,42 +86,46 @@ async function handleSimilarIssuesComment(context: Context, payload: IssuePayloa
...
       );
-      issueUrl.similarity = (issue.similarity * 100).toFixed(2);
+      issueUrl.similarity = Math.round(issue.similarity * 100).toString();
       return issueUrl;
     })
   );
 
-  const commentBody = issueList.map((issue) => `- [${issue.node.title}](${issue.node.url}) Similarity: ${issue.similarity}`).join("\n");
-  const body = `This issue seems to be similar to the following issue(s):\n\n${commentBody}`;
+  let finalIdx = 0;
```
0x4007, line 105: Never shorten words in identifiers to reduce cognitive overhead. I presume this is meant to mean `finalIndex`

src/handlers/issue-deduplication.ts
```diff
@@ -69,42 +86,46 @@ async function handleSimilarIssuesComment(context: Context, payload: IssuePayloa
...
+    .filter((issue) =>
+      matchRepoOrgToSimilarIssueRepoOrg(payload.repository.owner.login, issue.node.repository.owner.login, payload.repository.name, issue.node.repository.name)
+    )
+    .map((issue, index) => {
+      const modifiedUrl = issue.node.url.replace("https://github.com", "https://www.github.com");
+      finalIdx += 1;
+      return `[^0${index + 1}^]: [${issue.node.title}](${modifiedUrl}) ${issue.similarity}%`;
+    })
+    .join("\n");
+  const footnoteLinks = [...Array(finalIdx).keys()].map((i) => `[^0${i + 1}^]`).join("");
```
0x4007, line 116:
```suggestion
  const footnoteLinks = [...Array(++finalIndex).keys()].map((i) => `[^0${i + 1}^]`).join("");
```

src/handlers/issue-deduplication.ts
```diff
@@ -69,42 +86,46 @@ async function handleSimilarIssuesComment(context: Context, payload: IssuePayloa
...
-  const commentBody = issueList.map((issue) => `- [${issue.node.title}](${issue.node.url}) Similarity: ${issue.similarity}`).join("\n");
-  const body = `This issue seems to be similar to the following issue(s):\n\n${commentBody}`;
+  let finalIdx = 0;
+  const commentBody = issueList
+    .filter((issue) =>
+      matchRepoOrgToSimilarIssueRepoOrg(payload.repository.owner.login, issue.node.repository.owner.login, payload.repository.name, issue.node.repository.name)
+    )
+    .map((issue, index) => {
+      const modifiedUrl = issue.node.url.replace("https://github.com", "https://www.github.com");
+      finalIdx += 1;
```
0x4007, line 112:
```suggestion
```

src/handlers/issue-deduplication.ts
0x4007:
```suggestion
  let finalIndex = 0;
```

0x4007:
I made changes here but why is there no build CI?

0x4007 (approved):

0x4007:
Would be great if you can add this here https://github.com/ubiquity-os/plugin-template/blob/development/.github/workflows/build.yml

PR #39, opened 2024-10-17
fix: line break issue

### LINKED ISSUE #38: Line Breaks are getting removed
The relevance scoring editing messed up my formatting. This is not good. Im editing from mobile now

_Originally posted by @0x4007 in https://github.com/ubiquity/devpool-directory-tasks/issues/46#issuecomment-2418094146_

sshivaditya2019:
Resolves #38 

- Resolves the problem of line breaks being removed.

### DIFF
--- a/src/handlers/issue-deduplication.ts
+++ b/src/handlers/issue-deduplication.ts
@@ -40,6 +40,7 @@ export async function issueChecker(context: Context): Promise<boolean> {
   }
   issueBody = removeFootnotes(issueBody);
   const similarIssues = await supabase.issue.findSimilarIssues(issue.title + removeFootnotes(issueBody), context.config.warningThreshold, issue.node_id);
+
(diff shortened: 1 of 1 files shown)


src/handlers/issue-deduplication.ts
```diff
@@ -259,5 +259,5 @@ export function removeFootnotes(content: string): string {
       contentWithoutFootnotes = contentWithoutFootnotes.replace(new RegExp(`\\[\\^${footnoteNumber}\\^\\]`, "g"), "");
     });
   }
-  return contentWithoutFootnotes.replace(/\n{2,}/g, "\n").trim();
+  return contentWithoutFootnotes.replace(/\n{2,}/g, "\n");
```
0x4007, line 262: Why are you replacing line breaks at all? You should just leave them as is.

0x4007:
You should post QA but its such a small change so I'll let it slide

PR #41, opened 2024-10-26
fix: add caution message

### LINKED ISSUE #40: Improve UX for Match ( >90%) Similarity
If we have an issue with more than 90% match similarity, it should provide some information to the user instead of just closing the issue.

sshivaditya2019:
Resolves #40 

- Adds Caution Message to the Issue Body when its more than match similarity

QA: [Issue](https://github.com/sshivaditya2019/test-public/issues/191)

### DIFF
--- a/src/handlers/issue-deduplication.ts
+++ b/src/handlers/issue-deduplication.ts
@@ -201,7 +201,7 @@ async function handleMatchIssuesComment(
   const existingFootnotes = issueBody.match(footnoteRegex) || [];
   // Find the index with respect to the issue body string where the footnotes start if they exist
   const footnoteIndex = existingFootnotes[0] ? issueBody.indexOf(existingFootnotes[0]) : issueBody.length;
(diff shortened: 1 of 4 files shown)


src/handlers/issue-deduplication.ts
```diff
@@ -211,6 +178,72 @@ async function handleSimilarIssuesComment(
...
+
+  if (!issueBody) {
+    return;
+  }
+  // Find existing footnotes in the body
+  const footnoteRegex = /\[\^(\d+)\^\]/g;
+  const existingFootnotes = issueBody.match(footnoteRegex) || [];
+  // Find the index with respect to the issue body string where the footnotes start if they exist
+  const footnoteIndex = existingFootnotes[0] ? issueBody.indexOf(existingFootnotes[0]) : issueBody.length;
+  let resultBuilder = "\n\n>[!CAUTION]\n> This issue is very similar to the following issues:\n";
```
0x4007, line 204:
```suggestion
  let resultBuilder = "\n\n>[!CAUTION]\n> This issue may be a duplicate of the following issues:\n";
```

PR #44, opened 2024-11-09
fix: returnDataToKernel

### LINKED ISSUE #43: Actions are not working
All of the `compute.yml` runs are failing for this plugin which means that this plugin is not actually in use at all

Afaik we are using shivs DB and they are running a script to periodically update their DB with comments from the Ubiquity orgs. I don't know where that script is or what exactly it's fetching to replicate the DB which I need to be able to build https://github.com/ubiquity-os-marketplace/ubiquity-os-kernel-telegram/issues/17

shiv810:
Resolves #43 

- Previous failed runs still retained data, and the corresponding comment/issues should be present in the database. As far as I know, this isn’t running on either the organization’s supabase instance or mine.
- Fixed the returnDataToKernel, was missing `GITHUB_TOKEN` in compute.yml
- QA Action Runs: [Link](https://github.com/sshivaditya2019/issue-comment-embeddings/actions/runs/11754096399/job/32747697111)
- Good values for thresholds:
```js
matchThreshold: 0.9
warningThreshold: 0.7
jobMatchingThreshold: 0.7
```


src/handlers/issue-deduplication.ts
```diff
@@ -292,9 +285,15 @@ export function removeFootnotes(content: string): string {
       contentWithoutFootnotes = contentWithoutFootnotes.replace(new RegExp(`\\[\\^${footnoteNumber}\\^\\]`, "g"), "");
     });
   }
+  contentWithoutFootnotes = removeCautionMessages(contentWithoutFootnotes);
   return contentWithoutFootnotes;
 }
 
+export function removeCautionMessages(content: string): string {
+  const cautionRegex = />[!CAUTION]\n> This issue may be a duplicate of the following issues:\n((> - \[[^\]]+\]\([^)]+\)\n)+)/g;
```
0x4007, line 293: What are these checks? Seems like wrong approach to be doing this. I would expect you should be able to select the right data in the first place and process less comments with a better strategy, although I don't understand the full context of this process here.

src/adapters/supabase/helpers/comment.ts
```diff
@@ -28,11 +28,10 @@ export class Comment extends SuperSupabase {
     //First Check if the comment already exists
     const { data, error } = await this.supabase.from("issue_comments").select("*").eq("id", commentNodeId);
     if (error) {
-      this.context.logger.error("Error creating comment", error);
-      return;
+      throw new Error("Error creating comment");
     }
     if (data && data.length > 0) {
-      this.context.logger.info("Comment already exists");
```
0x4007, line 35: You should be able to just throw this, then it logs and halts function execution.

0x4007 (changes requested):

src/adapters/supabase/helpers/comment.ts
0x4007: @gentlementlegen rfc

0x4007: @gentlementlegen rfc

gentlementlegen: You can just `throw logger.fatal` so you get the log displayed and the error caught on a higher level of execution.

0x4007: Probably should be error

src/adapters/supabase/helpers/comment.ts
```diff
@@ -26,32 +26,41 @@ export class Comment extends SuperSupabase {
     issueId: string
   ) {
     //First Check if the comment already exists
-    const { data, error } = await this.supabase.from("issue_comments").select("*").eq("id", commentNodeId);
-    if (error) {
-      this.context.logger.error("Error creating comment", error);
+    const { data: existingData, error: existingError } = await this.supabase.from("issue_comments").select("*").eq("id", commentNodeId);
+    if (existingError) {
+      this.context.logger.error("Error creating comment");
```
gentlementlegen, line 31: Maybe you can also add the error itself in the log for clarity.

src/handlers/add-issue.ts
```diff
@@ -8,27 +8,28 @@ export async function addIssue(context: Context) {
     adapters: { supabase },
   } = context;
   const { payload } = context as { payload: IssuePayload };
-  const markdown = payload.issue.body + " " + payload.issue.title || null;
-  const authorId = payload.issue.user?.id || -1;
-  const nodeId = payload.issue.node_id;
+  const issue = payload.issue;
+  const markdown = issue.body + " " + issue.title || null;
```
gentlementlegen, line 12: This is confusing. Use a string template or add parenthesis.

gentlementlegen:
@sshivaditya2019 Did you retry with the newly added SDK?

gentlementlegen:
Yes it is optional but you will get the invalid signature error if it doesn't match the kernel in use. In this org it's fine because it is defined for dev and production. Locally you can disable it if you need to test.

gentlementlegen:
You should use `ncc` locally and try to run it without minifying so you can see the error clearly. A guess is that the plugin is compiled as ESM when it should be CJS or vice versa.

.github/workflows/formatting-checks.yml
```diff
@@ -12,6 +12,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Setup Node
```
gentlementlegen, line 15: I don't think this is needed since `bun` is there.

package.json
```diff
@@ -17,7 +17,7 @@
     "knip-ci": "knip --no-exit-code --reporter json --config .github/knip.ts",
     "prepare": "husky install",
     "test": "cross-env NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" jest --setupFiles dotenv/config --coverage",
-    "worker": "wrangler dev --env dev --port 4004",
+    "worker": "wrangler dev --env dev --port 4000",
```
gentlementlegen, line 20: I know that @whilefoo had set 4004 on purpose.

src/handlers/issue-matching.ts
```diff
@@ -41,44 +42,53 @@ export async function issueMatching(context: Context<"issues.opened" | "issues.e
...
+  });
   if (similarIssues && similarIssues.length > 0) {
-    similarIssues.sort((a, b) => b.similarity - a.similarity); // Sort by similarity
-    const fetchPromises = similarIssues.map(async (issue) => {
-      const issueObject: IssueGraphqlResponse = await context.octokit.graphql(
-        `query ($issueNodeId: ID!) {
+    similarIssues.sort((a: IssueSimilaritySearchResult, b: IssueSimilaritySearchResult) => b.similarity - a.similarity); // Sort by similarity
+    const fetchPromises = similarIssues.map(async (issue: IssueSimilaritySearchResult) => {
+      try {
+        const issueObject: IssueGraphqlResponse = await context.octokit.graphql(
```
gentlementlegen, line 54:
```suggestion
        const issueObject: IssueGraphqlResponse = await context.octokit.graphql(/* GraphQL */
```

gentlementlegen: It should, if I didn't misspell it. Your IDE should understand the string is meant to be gql code and highlight / prettify.

gentlementlegen: You should also probably add: https://github.com/ubiquity-os-marketplace/text-conversation-rewards/blob/4de373fc13fea8fc18c7d016379ad5b909321b87/graphql.config.yml

gentlementlegen:
@sshivaditya2019 tested, works on my side: https://github.com/Meniole/text-vector-embeddings/actions/runs/12063952321/job/33639960768
Can you please just add the gql file I recommended then good to merge.

PR #54, opened 2024-11-30
Always Recommend Contributors

### LINKED ISSUE #45: Always Recommend Talent
We need to calibrate the relevance thresholds for issue dedupe and the talent recommendation because i havent seen them work since we first deployed the plugins.

I think we should always suggest at least one contributor (three default, in order of relevance) so we can test and see how accurate it is

This should be a config option to always recommend 
```yml
alwaysRecommend: number
```
If its 0 or missing we can do as we are.

shiv810:
Resolves #45

1. The DB gets populated by handling issue complete webhook
2. `alwaysRecommends` the specified number of contributors if not specified config works as usual.


0x4007:
QA?

0x4007:
Why does it only work after you close it? It's supposed to be immediate after the issue is opened or edited.

0x4007:
> > Why does it only work after you close it? It's supposed to be immediate after the issue is opened or edited.
> 
> In the QA examples, all of them had similarity more than 90%. Otherwise, it would work when edited and labeled
> 
> [Example warningThreshold](https://github.com/sshivaditya2019/test-public/issues/233)

Sorry but this doesn't make any sense to me. It doesn't matter what the similarity level is, it should always recommend talent, as per the title of the task.

0x4007 (approved):
QA seems fine

PR #59, opened 2024-12-08
feat: adds scraper script

### LINKED ISSUE #56: Scraper: Populate "Closed As Complete" Issue Specifications
We need to populate our matchmaking database. We "register" users in our system when they "Sign in with GitHub" on https://devpool.directory/

I suppose with GraphQL it may be simple to check every issue that is closed as complete with these assignees. 

- We need to capture all of the metadata to make it easier to work with later.
- We need to be able to determine the location of the issue, which means organization and repository. We should capture both the names as well as the IDs, in case of renames. 
- We need to generate vector embeddings and store the specifications in plain text. 

For now we can work off of a static snapshot of registrants. In the future, we should automatically scrape their completed issues as soon as they register on our site. 

[auth.users.json](https://github.com/user-attachments/files/18005367/auth.users.json)
[auth.users.csv](https://github.com/user-attachments/files/18005368/auth.users.csv)

sshivaditya2019:
Resolves #56 

- [X] Scrapes issues based on the `username` passed in.
- [X] Reads the token either as a user input or from the cli.
- [X] Updates issues in the repo, even with same node_id exists. 
- [x] Issue Dedup and Matchmaking Results.

### DIFF
--- a/.github/knip.ts
+++ b/.github/knip.ts
@@ -3,7 +3,14 @@ import type { KnipConfig } from "knip";
 const config: KnipConfig = {
   entry: ["src/main.ts"],
   project: ["src/**/*.ts"],
-  ignore: ["src/types/config.ts", "**/__mocks__/**", "**/__fixtures__/**", "src/types/database.ts", "src/handlers/user-issue-scraper.ts"],
+  ignore: [
+    "src/types/config.ts",
+    "**/__mocks__/**",
+    "**/__fixtures__/**",
+    "src/types/database.ts",
+    "src/handlers/user-issue-scraper.ts",
+    "src/handlers/issue-scraper.ts",
+  ],
   ignoreExportsUsedInFile: true,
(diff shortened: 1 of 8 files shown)


0x4007:
Yes and please update the database with it. You can QA with some task matchmaking scoring improvements and second goal some issue dedupe improvements- right?

0x4007:
Would you mind QA testing using real issues or is that too complex to set up? 

You could, for example, output results in console instead of actually posting them across our real issues. 

Based on those results we can know for sure how this is expected to perform after we merge it in?

0x4007:
I looked through a few of the results and I think above 80% seems actually relevant. What are your thoughts? 

It might make sense to exclude showing matches below 80%?

0x4007:
@sshivaditya @sshivaditya2019 you can merge now?

0x4007 (approved):
Passes the CI so it's fine

PR #70, opened 2025-01-17
feat: implement annotate command

### LINKED ISSUE #67: `/annotate`
There are situations where we write about certain issues that were likely to have been posted in the past, but are difficult to find. 

Imagine if using vector embeddings, we can instantly link to whatever issues are being referenced: 

> We had some problems with KV, so I don't know if it's wise to enable plugin communication without fixing the problem (I can't find the issue about this). Alternatively we could make this as part of a module inside  `text-conversation-rewards`

From https://github.com/kingsley-einstein/contributions-scan/pull/4#issuecomment-2577468623

Using this command, we should replicate the behavior of issue deduplication on the existing comment and edit/add the footnotes with links to the source issues.

A limitation that I see is how wide of a search we should conduct. Within the same repository is somewhat useless, and globally might incur too much noise. 

Organization wide might be the best default, but sometimes it would be very useful, especially for us using three organizations, to do a global annotation/search. 

Perhaps we can have optional arguments to scope the search.

```
/annotate https://github.com/kingsley-einstein/contributions-scan/pull/4#issuecomment-2577468623 global
```

But default can be just

```
/annotate
```

And it will automatically annotate the previous comment with an organization wide search.

koya0:
Resolves #67 

## QA
image


koya0:
i was going to add the annotate command to manifest.json, but when kernel identifies the command object in manifest and the comment made doesnt start with the command, it skips the plugin

koya0:
i commited bun.lockb in [feat: create Comment type](https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/70/commits/f1164c152bf05e9c575a88f5926507add8cc5c64), but it was conflicting with the old one, so I updated the file to the old one in [chore: update bun.lockb](https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/70/commits/708fae554aedaa28381da933d15f6c9684e3be96), but now the tests are failing

0x4007:
QA screenshot looks promising but better if you can link that thread and we can try ourselves.

src/handlers/annotate.ts
```diff
@@ -0,0 +1,157 @@
+import { Context } from "../types";
+import { Comment } from "../types/comment";
+import { processSimilarIssues, IssueGraphqlResponse } from "./issue-deduplication";
+
+export async function annotate(context: Context, commentId: string, scope: string) {
+  const { logger, octokit, payload } = context;
+
+  const repository = payload.repository;
+
+  if (commentId === "") {
```
0x4007, line 10: Wouldn't it make more sense to have a null comment ID

src/handlers/annotate.ts
```diff
@@ -0,0 +1,157 @@
...
+ * @returns True if a annotate footnote exists, false otherwise
+ */
+export function checkIfAnnotateFootNoteExists(content: string): boolean {
+  const footnoteDefRegex = /\[\^(\d+)\^\]: \d+% similar to issue: [^\n]+(\n|$)/g;
+  const footnotes = content.match(footnoteDefRegex);
+  return !!footnotes;
+}
+
+/**
+ * Removes all footnotes from the comment content.
```
0x4007, line 141: This seems problematic if a comment contains normal footnotes. 

As I recall we identified deduplication footnotes because they are prefixed with 0. 

Normally usage of footnotes is just a count from 1 but we learned that GitHub just requires a unique ID and it doesn't have to be a number. 

So we prefix deduplication footnotes with 0 and count up like

```
01
02
03
```

src/handlers/user-annotate.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  if (!context.command) {
+    return;
+  }
+
+  if (context.command.name === "annotate") {
+    const commentUrl = context.command.parameters.commentUrl ?? "";
+    const scope = context.command.parameters.scope ?? "org";
+    if (commentUrl !== "") {
```
0x4007, line 18: Maybe null again. Empty strings seem to be LLM code smell

src/handlers/user-annotate.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+export async function userAnnotate(context: Context<"issue_comment.created">) {
+  const { logger } = context;
+  const comment = context.payload.comment;
+  const splitComment = comment.body.trim().split(" ");
+  const commandName = splitComment[0].replace("/", "");
+
+  let commentId = "";
```
0x4007, line 36: Null or undefined makes more sense.

0x4007 (changes requested):

koya0:
> QA screenshot looks promising but better if you can link that thread and we can try ourselves.

https://github.com/koya0/.ubiquity-os/issues/22

0x4007:
> i was going to add the annotate command to manifest.json, but when kernel identifies the command object in manifest and the comment made doesnt start with the command, it skips the plugin

@gentlementlegen @whilefoo RFC

src/handlers/annotate.ts
koya0: sorry, i guess the comment is wrong, it should be "Removes annotate footnotes from the comment content.". This function is basically the `removeFootnotes` function from `issue-deduplication.ts` but the regex is changed to match the annotate footnote format

gentlementlegen:
It will skip the plugin only for `issue.comment_created` scenario, although I guess this plugin relies on it to do the embedding as well?

0x4007:
This will look up from our already generated embeddings. It needs to only respond to the command.

gentlementlegen:
@0x4007 It uses this event to store the comment in the database: https://github.com/ubiquity-os-marketplace/text-vector-embeddings/blob/0bacd3c4a2099d13261a9d8154160dc12281ba36/src/plugin.ts#L37 which is why it conflicts with the command.

0x4007:
So what's the solution

whilefoo:
The kernel should be modified so that if the plugin manifest defines a command but empty `ubiquity:listeners`, it should only run the plugin if the command matches, but if `ubiquity:listeners` also includes `issue_comment.created` it should be run on any comment

0x4007:
Given this task I think we should support commands in the copilot chat box instead: 

https://github.com/ubiquity-os/plugins-wishlist/issues/54

whilefoo:
> Given this task I think we should support commands in the copilot chat box instead:
> 
> [ubiquity-os/plugins-wishlist#54](https://github.com/ubiquity-os/plugins-wishlist/issues/54)

This is an additional way to trigger commands, right? You'd still be able to use commands in the comments? Or we would have different set of commands for the comments (start, stop) and different set for chatbox?

whilefoo:
> @0x4007 do you have access to it on mobile devices? I don't seem to have it, on android.

I think on mobile it's only for Github Pro users

koya0:
> @koya0, this task has been idle for a while. Please provide an update.
> 
> 
> 

Waiting for review

koya0:
> @koya0 You have to update `bun.lockb` because it is not up to date and breaks the workflows.

https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/70#issuecomment-2599321605

I will try again

gentlementlegen (commented):
The code looks generally okay, for the lock file I think you should merge the `development` branch here, then `bun install` and commit all the changes.

Could you add unit tests related to the annotation and the modification of the comment since this is a new feature?

koya0:
> The code looks generally okay, for the lock file I think you should merge the `development` branch here, then `bun install` and commit all the changes.
> 
> Could you add unit tests related to the annotation and the modification of the comment since this is a new feature?

I tried what you said but it didnt work. I cant even resolve the merge conflict, because I cant open merge editor and when try to accept current changes or incoming changes, nothing happens. I suppose its because is a binary file, and github probably doesnt have much support for that

gentlementlegen:
@koya0 You can try upgrading bun to `1.2` where lock files became plain text files which should help you merging then.

0x4007:
I had a similar problem with bun I needed to upgrade it before it started generating compatible lock files.

whilefoo:
Normally if there's a conflict with `bun.lockb`, you just need to run `bun install` and it will generate a new lockfile

koya0:
I updated bun, but still couldnt resolve the conflicts because althought my file was plain text the other one was binary. What I did was to replace my bun.lockb with bun.lockb from upstream and use the plain text bun.lock from the upgraded bun

gentlementlegen:
@koya0 Can't you just delete it now that you are using the plain text one?

koya0:
i tried this i [chore: generate plain text bun lock file](https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/70/commits/e1b87ffe5b560eb0a316f59176a9ef5f3421a30e) but 

> @koya0 Can't you just delete it now that you are using the plain text one?

i deleted it in [chore: generate plain text bun lock file](https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/70/commits/e1b87ffe5b560eb0a316f59176a9ef5f3421a30e) but it still had conflicts. But i deleted now and worked, lol

0x4007:
> > @0x4007 do you have access to it on mobile devices? I don't seem to have it, on android.
> 
> 
> 
> I think on mobile it's only for Github Pro users

https://docs.github.com/en/copilot/using-github-copilot/asking-github-copilot-questions-in-github-mobile

Looks like we need to pay for GitHub copilot so this can be a problem. Maybe we should port to the telegram chatbot instead.

0x4007:
I realized also that we could do a workaround by using the web browser from mobile instead. It's available there as well. image

gentlementlegen:
I confirm it is available in the browser on mobile versions. But when I try to use it I get the following screen:
image

So I don't think this is reliable, we need to find a workaround for the problem related to this task in the meantime.

0x4007:
Then comment slash command for now

src/handlers/annotate.ts
```diff
@@ -0,0 +1,157 @@
...
+
+async function handleSimilarIssuesComment(
+  context: Context,
+  payload: Context["payload"],
+  commentBody: string,
+  commentId: number,
+  issueList: IssueGraphqlResponse[]
+) {
+  // Find existing footnotes in the body
+  const footnoteRegex = /\[\^(\d+)\^\]/g;
```
0x4007, line 93: Maybe we should prefix the footnotes for issue de duplication with a special character to target them more precisely @shiv810 rfc

src/handlers/annotate.ts
```diff
@@ -0,0 +1,157 @@
...
+  const existingFootnotes = commentBody.match(footnoteRegex) || [];
+  const highestFootnoteIndex = existingFootnotes.length > 0 ? Math.max(...existingFootnotes.map((fn) => parseInt(fn.match(/\d+/)?.[0] ?? "0"))) : 0;
+  let updatedBody = commentBody;
+  let footnotes: string[] | undefined;
```
0x4007, line 97:
```suggestion
  let footnotes: string[];
```

src/handlers/annotate.ts
```diff
@@ -0,0 +1,157 @@
...
+  // Sort relevant issues by similarity in ascending order
+  issueList.sort((a, b) => parseFloat(a.similarity) - parseFloat(b.similarity));
+  issueList.forEach((issue, index) => {
+    const footnoteIndex = highestFootnoteIndex + index + 1; // Continue numbering from the highest existing footnote number
+    const footnoteRef = `[^0${footnoteIndex}^]`;
```
0x4007, line 102: I don't know off hand if 10 turns into 010 or just 10

src/handlers/annotate.ts
```diff
@@ -0,0 +1,157 @@
...
+  });
+}
+
+/**
+ * Checks if a annotate footnote exists in the content.
+ * @param content The content to check for annotate footnotes
+ * @returns True if a annotate footnote exists, false otherwise
+ */
+export function checkIfAnnotateFootNoteExists(content: string): boolean {
+  const footnoteDefRegex = /\[\^(\d+)\^\]: \d+% similar to issue: [^\n]+(\n|$)/g;
```
0x4007, line 135: This seems brittle to target strings like this. Any ideas to get more reliable cross plugin output targeting? 

@Shiv810 @gentlementlegen @whilefoo rfc

0x4007 (approved):

src/handlers/annotate.ts
koya0: should I make this change in `issue-deduplication.ts` too?

0x4007: Sure. Just seems more concise to write it that way

src/handlers/annotate.ts
0x4007: Can you make it?

0x4007:
/ask do the changes include updating the manifest for this plugin to explain how to use the command and have a provided example? 

Our plugins invoked by commands should populate on our help menu. The help menu reads all that information from the compiled manifest files which are derived from some typescript source code.

0x4007:
Actually I see that the manifests aren't building on push @gentlementlegen what should @koya0 do here?

0x4007:
> A task appears to be needed to ensure the manifests are building properly on push to provide this documentation.

Agreed on the recommendation. @gentlementlegen maybe you can handle this task. For now, perhaps the manifest.json can be manually updated within this pull @koya0

gentlementlegen:
It is normal that the manifest is not built, there are no push events on our repo during a pull-request because the push happens on the user's repository not ours. The user just needs to enable the workflows on its side and the manifest should update, otherwise it will update once we merge inside of our repository.

bun.lockb
0x4007, line 1: I suppose this should be removed.

0x4007 (approved):
My review status doesn't appear to be approved so approving one more time because this has long been ready to merge in my opinion.

0x4007:
@shiv810 @whilefoo @gentlementlegen you guys can do a retroactive review because this has been pending for too long.

PR #73, opened 2025-01-23
fix: null handling for graphql search result

### LINKED ISSUE #72: null result from GraphQL search
> ```
> 
> 
> 
> ```
> 
> @shiv810  again here
>  

 _Originally posted by @0x4007 in [#247](https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/247#issuecomment-2610775622)_

shiv810:
Resolves #72

- Adds type check for the graphql result and skips if null in `issue-contributor matchmaking`


manifest.json
```diff
@@ -33,5 +33,5 @@
       }
     }
   },
-  "homepage_url": "https://ubiquity-os-comment-vector-embeddings-development.ubiquity.workers.dev"
+  "homepage_url": "https://ubiquity-os-comment-vector-embeddings-development.sshivaditya.workers.dev"
```
0x4007, line 36: Perhaps this should be changed back

src/handlers/issue-matching.ts
```diff
@@ -98,7 +98,11 @@ export async function issueMatching(context: Context<"issues.opened" | "issues.e
       if (!issuePromise || issuePromise.status === "rejected") {
         return;
       }
-      const issue = issuePromise.value as IssueGraphqlResponse;
+      const issue = issuePromise.value;
```
0x4007, line 101: What about the type?

manifest.json
0x4007: Sure but `sshivaditya` came from somewhere - is it your secrets? @gentlementlegen rfc

gentlementlegen: The action will update this once merge to our organization automatically.

0x4007 (approved):

PR #84, opened 2025-02-07
feat: implement annotate config

### LINKED ISSUE #79: Annotate config
It should have similar config options as deduplication but they should be unique and only for annotate feature. 

Rename the config properties for deduplication and annotate to be sensible. Probably give them similar names but prefix with dedupe or annotate. 

`warning.Threshold` from issue deduplication? It should have its own config properties surely! We should make another task for that.

_Originally posted by @0x4007 in https://github.com/ubiquity-os-marketplace/text-vector-embeddings/issues/67#issuecomment-2637993361_

koya0:
Resolves #79


koya0:
maybe the default threshold should be smaller, like 0.3, not sure

koya0:
@0x4007 can you please review?

tests/main.test.ts
```diff
@@ -621,6 +621,7 @@ describe("Plugin tests", () => {
         warningThreshold: 0.75,
         matchThreshold: 0.95,
```
0x4007, line 622: For further disambiguation we should also rename these to reference dedupe

0x4007:
You should also show QA after these renames

koya0:
> You should also show QA after these renames

Sorry, I tested but didn't link the QA: https://github.com/koya0/.ubiquity-os/issues/38
Now issues with less than 75% are being annotated, the only problem is that the issues with 63% similarity are not very similar to the comment, I guess the supabase function Im using to find similar issues isnt very good.

0x4007:
I might be mixing this up with a different algorithm but for task matchmaking anything above 80% match is pretty accurate. @shiv810 mentioned something about the normalization peaking at around 80% due to the specific types of embeddings we use but the details are beyond me. 

Anyways, if it's the same algorithm then above 80% is usually more useful.

koya0:
To get similar issues for the annotate comment, you want to consider, besides the issue body and title, every comment made in the issue? First thing that comes to mind is to make another column in the issues table that stores the issue body, title and all the comments concatenated and make a embedding for this column.

0x4007:
> That should be fairly simple, we would have to write a new pg function to search for similar comments, like how its done in command-ask

- shiv810 via telegram

koya0:
But then the comments are going to be treated independently of the issue spec, right? So if we find a issue comment with 84% similarity, but the issue spec is only 10% similar to the annotate comment, we would add a footnote that the comment is 84% similar to the issue? Or should make an average similarity between issue comment and issue spec?

0x4007:
Just link to comment

koya0:
now similar comments are being added to the footnotes: https://github.com/koya0/.ubiquity-os/issues/38

0x4007:
interesting result I just wish it wouldn't annotate the same sentence multiple times. 

Might be out of scope for this task but any ideas for fixes there

koya0:
> interesting result I just wish it wouldn't annotate the same sentence multiple times. 
> 
> Might be out of scope for this task but any ideas for fixes there 

The command annotates the sentence that is most similar to the issue/comment. I don't know how could we change this feat to the same sentence dont be annotated multiple times.

koya0:
Build a scraper that creates a JSONL compatible dataset, following OpenAI's spec for the fine-tune dataset, to be used with their framework.

We need close to 250 to 300 well annotated and clean examples, for the training set. And a separate validation set which should have another 100 to 150 examples.

koya0:
/annotate https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/84#issuecomment-2660070322 org

koya0:
.

koya0:
@shiv810 this is the current find_similar_issues_ function prod supabase is using: https://github.com/ubiquity-os-marketplace/text-vector-embeddings/blob/development/supabase/migrations/20241026185200_function_issue.sql, right? With annotate, when I use find_similar_issues, I use current_id as the comment_id so basically Im not getting correctly
`payload->'repository'->>'name'::text`
`payload->'repository'->'owner'->>'login'::text`
I think this is potentially the problem annotate is having you can see in these logs that the similar issues the command is getting are suspicious:
https://github.com/ubiquity-os-marketplace/text-vector-embeddings/actions/runs/13336709064/job/37253457236
https://github.com/ubiquity-os-marketplace/text-vector-embeddings/actions/runs/13324461755/job/37214760918#step:4:1
https://github.com/ubiquity-os-marketplace/text-vector-embeddings/actions/runs/13335803573/job/37250610875

koya0:
I think we have to use another supabase function only for annotate

0x4007:
Sounds like https://github.com/ubiquity-os-marketplace/text-vector-embeddings/pull/84#issuecomment-2646546859

koya0:
> @koya0, this task has been idle for a while. Please provide an update.
> 
> 
> 

Waiting for review

shiv810:
> @shiv810 this is the current find_similar_issues function prod supabase is using: https://github.com/ubiquity-os-marketplace/text-vector-embeddings/blob/development/supabase/migrations/20241026185200_function_issue.sql, right? With annotate, when I use find_similar_issues, I use current_id as the comment_id so basically Im not getting correctly `payload->'repository'->>'name'::text` `payload->'repository'->'owner'->>'login'::text` I think this is potentially the problem annotate is having. 

The function's outputs are expected, as we're filtering for results within a specific repository in that particular function. I would suggest creating your own function to handle annotations instead.

koya0:
> > @shiv810 this is the current find_similar_issues function prod supabase is using: https://github.com/ubiquity-os-marketplace/text-vector-embeddings/blob/development/supabase/migrations/20241026185200_function_issue.sql, right? With annotate, when I use find_similar_issues, I use current_id as the comment_id so basically Im not getting correctly `payload->'repository'->>'name'::text` `payload->'repository'->'owner'->>'login'::text` I think this is potentially the problem annotate is having.
> 
> The function's outputs are expected, as we're filtering for results within a specific repository in that particular function. I would suggest creating your own function to handle annotations instead.

where should I commit the functions?

0x4007:
@shiv810  maybe you can help add it to Supabase. 

Also probably makes sense to save Supabase functions in some directory in this repository so we can review and such

koya0:
> @koya0, this task has been idle for a while. Please provide an update.
> 
> 
> 

Waiting for @shiv810

shiv810:
> https://github.com/ubiquity-os-marketplace/text-vector-embeddings/blob/development/supabase/migrations/20241026185200_function_issue.sql

The functions should be in the migrations dir, and on merge, they would be added to Supabase directly.

koya0:
> > > @koya0, this task has been idle for a while. Please provide an update.
> > 
> > Waiting for @shiv810
> 
> The functions should be in the migrations dir, and on merge, they would be added to Supabase directly.

How do you create the file's name with timestamp for the migration?

shiv810:
> How do you create the file's name with timestamp for the migration?

You should refer to the [supabase migration guide](https://supabase.com/docs/guides/deployment/database-migrations)

koya0:
I made some tests in https://github.com/koya0/.ubiquity-os/issues/41 and defined a 0.7 weight for the cosine_distance and 0.3 for the l2_distance and corrected the annotate threshold to 0.65

koya0:
@shiv810 can you please review?

shiv810 (approved):

src/handlers/annotate.ts
```diff
@@ -122,13 +187,57 @@ async function handleSimilarIssuesComment(
...
+}
+
 /**
  * Checks if a annotate footnote exists in the content.
  * @param content The content to check for annotate footnotes
  * @returns True if a annotate footnote exists, false otherwise
  */
 export function checkIfAnnotateFootNoteExists(content: string): boolean {
-  const footnoteDefRegex = /\[\^(\d+)\^\]: \d+% similar to issue: [^\n]+(\n|$)/g;
+  const footnoteDefRegex = /\[\^(\d+)\^\]: \d+% similar to (issue|comment): [^\n]+(\n|$)/g;
```
0x4007, line 240: I still think that the regex is so shaky. If its based on the output of another plugin, this will lead to hard to diagnose bugs in the future.

src/handlers/annotate.ts
```diff
@@ -140,7 +249,7 @@ export function checkIfAnnotateFootNoteExists(content: string): boolean {
  * @returns The content without footnotes
  */
 export function removeAnnotateFootnotes(content: string): string {
-  const footnoteDefRegex = /\[\^(\d+)\^\]: \d+% similar to issue: [^\n]+(\n|$)/g;
+  const footnoteDefRegex = /\[\^(\d+)\^\]: \d+% similar to (issue|comment): [^\n]+(\n|$)/g;
```
0x4007, line 252: Here as well, potentially.

0x4007 (approved):

PR #86, opened 2025-02-09
fix: add dedupe timeout

### LINKED ISSUE #27: Similar Issue Editing Specification Timeout
I generally write specifications, post, edit, post, edit etc.

While I am editing, the script runs again and changes my specification body. This is a problem because when I complete my edit, GitHub refuses to accept my changes because there were other changes that occurred in the background from the bot. It tells me I need to refresh the page, and try again. 

This is obviously annoying so we need to make a timeout for edit. The timeout feature should be mindful of two things:
1. It should restart the timer if there is another edit. 
2. It should be configurable, with the default being 30 minutes.

shiv810:
Resolves #27 

- Issue Dedupe handler would sleep by default for `config.timeout` value, for all events.
- After, the sleep it fetches the latest issue, and checks for the update and the created timestamps, if any edits after the original timestamp the current event/workflow call is terminated
- By default the timeout is 30 mins


0x4007:
Weird this only happens for your pulls. Maybe something to do with private organization membership @gentlementlegen @whilefoo rfc

whilefoo:
> Weird this only happens for your pulls. Maybe something to do with private organization membership @gentlementlegen @whilefoo rfc

It seems it's not only for @shiv810 , I've just tested and also got the same [error](https://github.com/ubiquity-os-marketplace/text-conversation-rewards/pull/265)...I'm trying to figure out what's wrong

shiv810:
QA:

[Issue](https://github.com/ShivTestOrg/test-public/issues/257)

- [Opened](https://github.com/ShivTestOrg/issue-comment-embeddings/actions/runs/13297693028/job/37133220270) and followed by [edit](https://github.com/ShivTestOrg/issue-comment-embeddings/actions/runs/13297695038/job/37133225165) with the timeout period
- [Edit](https://github.com/ShivTestOrg/issue-comment-embeddings/actions/runs/13297709793/job/37133270621) followed by one more [edit](https://github.com/ShivTestOrg/issue-comment-embeddings/actions/runs/13297711245/job/37133275192)  within the timeout period.

src/handlers/issue-deduplication.ts
```diff
@@ -30,16 +77,46 @@ export async function issueChecker(context: Context<"issues.opened" | "issues.ed
...
+  if (lastEditedAt === null) {
+    logger.info("Last edited time is null, skipping deduplication check", {
+      issueNumber: originalIssue.number,
+    });
+    return;
+  }
+
+  const latestEventTime = new Date(lastEditedAt).getTime();
+
+  // Event should only proceed if it's the most recent action
```
0x4007, line 99: This is interesting how it's implemented have you tested that this works as expected?

0x4007 (approved):
Looks fine.

src/handlers/issue-deduplication.ts
shiv810: > This is interesting how it's implemented have you tested that this works as expected?

It does work. [Issue](https://github.com/ShivTestOrg/test-public/issues/257),  it skip issue-dedupe if its not the latest event specifically on the issue spec.

0x4007:
Time to test in prod. 

QA and testing for anything time based is always sketchy.

PR #105, opened 2025-07-19
feat: batch update for issue bodies

### LINKED ISSUE #95: New Sleep Strategy
The original problem to be solved is that when editing specifications, the dedupe would immediately jump in and add changes. This is a problem because while editing I was unable to save my changes because other changes occurred (dedupe details.)

We need another solution. 

> As a side note I feel like "sleeping" for 30 minutes is a huge problem for scaling, because we can only run six actions concurrently on any given repository. If we have six people that edited a spec in the last half hour, then no more vector embeddings can run. 
> 
> https://github.com/ubiquity-os-marketplace/text-vector-embeddings/actions/runs/13971522410/job/39114321487#step:4:140
> 
> We should come up with a workaround. @gentlementlegen @shiv810 rfc 

 _Originally posted by @0x4007 in [#94](https://github.com/ubiquity-os-marketplace/text-vector-embeddings/issues/94#issuecomment-2740647285)_

gentlementlegen:
Resolves #95


package.json
```diff
@@ -18,6 +18,7 @@
     "prepare": "husky install",
     "test": "cross-env NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" jest --setupFiles dotenv/config --coverage",
     "worker": "wrangler dev --env dev --port 4004",
+    "dev": "bun run --port 4004 --no-clear-screen --watch src/worker.ts",
```
0x4007, line 21: Consider deploying to Deno if we have problems with Workers.

gentlementlegen: No problem with workers, I prefer `bun` to run it because it loads the `.env` file instead of `.dev.vars` which is more convenient, and we kinda dropped workers anyway.

.github/workflows/compute.yml
```diff
@@ -32,21 +32,24 @@ jobs:
...
 
-      - name: setup node
-        uses: actions/setup-node@v4
-        with:
-          node-version: "20.10.0"
+      - uses: denoland/setup-deno@v2
 
       - name: Execute plugin
-        run: node dist/index.js
+        run: deno run --unstable-sloppy-imports --allow-all --unstable-kv dist/index.js
```
0x4007, line 45: Maybe simple enough to find and replace the source code and just make it deno native.

.github/workflows/jest-testing.yml
```diff
@@ -25,9 +21,86 @@ jobs:
...
 
-      - name: Jest With Coverage
-        run: bun run test
+      - name: Bun Test With Coverage
+        run: bun test --coverage --reporter=junit --reporter-outfile=./junit.xml
+
+      - name: Generate Markdown Dashboard from JUnit XML
+        run: |
+          bun add fast-xml-parser@4.2.5
+          cat <<'EOF' > junit-to-md.ts
```
0x4007, line 30: Seems like a strange approach you should just commit this file and run it instead of generating it on the fly every time. Also easier to test my way.

src/cron/workflow.ts
0x4007, line 1: Deno Cron?

0x4007 (changes requested):

.github/workflows/compute.yml
gentlementlegen: This runs what is compiled by `ncc` to fast run in Actions. The idea was just the replace `node` as the runner to `deno` so we have access to `KV` capabilities. There is no other way to run this with Deno except hosting the whole plugin itself within Deno. Even if made native the only change would be the `--unstable-sloppy-imports` removed so I don't think it's worth.

.github/workflows/jest-testing.yml
gentlementlegen: Why would we commit the results of a test run? The idea is to show this within the summary of the Action run. The problem is that `bun` does not offer markdown support like `jest` so I have to transform the output.

Result: https://github.com/ubiquity-os-marketplace/text-vector-embeddings/actions/runs/16502544944/attempts/1#summary-46664775389

whilefoo: Interesting, because [Bun docs](https://bun.com/docs/cli/test#github-actions) state:

> `bun test` automatically detects if it's running inside GitHub Actions and will emit GitHub Actions annotations to the console directly.
> 
> No configuration is needed, other than installing bun in the workflow and running `bun test`.

I think because you're explicitly adding `--reporter=junit --reporter-outfile=./junit.xml` you prevent bun from automatically outputting the correct format for Actions, maybe try it out

gentlementlegen: @whilefoo you are correct it can annotate, but I think it won't be capable of posting the annotations due to the fact that it happens on a fork, where the `GITHUB_TOKEN` is not allowed to annotate within the pull-request. The point of using the summary is that no special elevation is needed to post content there. I'll give it a try

.github/workflows/deno-deploy.yml
whilefoo, line 1: isn't this Actions-only plugin? why are we deploying it to Deno?

.github/workflows/cron.yml
```diff
@@ -0,0 +1,47 @@
...
+  workflow_dispatch:
+  schedule:
+    - cron: "0 * * * *"
+
+env:
+  APP_ID: ${{ secrets.APP_ID }}
+  APP_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}
+  APP_INSTALLATION_ID: ${{ secrets.APP_INSTALLATION_ID }}
+  DENO_KV_ACCESS_TOKEN: ${{ secrets.DENO_DEPLOY_TOKEN }}
+  DENO_KV_URL: ${{ secrets.DENO_KV_URL }}
```
whilefoo, line 13: is `DENO_KV_ACCESS_TOKEN` used implicitly by Deno to access the KV?

src/cron/database-handler.ts
```diff
@@ -0,0 +1,67 @@
+import { parseGitHubUrl } from "../helpers/github";
+
+export const KV_PREFIX = "cron";
+
+export class CronDatabase {
+  private _kv: Deno.Kv;
+
+  constructor(kv: Deno.Kv) {
+    this._kv = kv;
+  }
```
whilefoo, line 10: `get` method is eventually consistent by default which means it can return stale values and at every cron run we loop through the issues, update the issue and remove them from the DB. 
Since each remove method fetches from the KV and then sets it, you can get stale value which doesn't have the issue from previous loop iteration removed.
Fortunately for us Deno supports reading in strong consistency mode with `{ consistency: "strong" }`

gentlementlegen: I could change this using `atomic` so we do it in one commit only. Maybe this would also benefit for the quota, I can look into that

.github/workflows/deno-deploy.yml
gentlementlegen: You need a project in Deno because each project exposes its own url for the database. Maybe only creating the project without deploying would be sufficient, I just thought using our action would streamline the process if we ever need to actually have a Deno endpoint available

src/cron/index.ts
```diff
@@ -0,0 +1,95 @@
+import { createAppAuth } from "@octokit/auth-app";
+import { Octokit } from "@octokit/rest";
+import { customOctokit } from "@ubiquity-os/plugin-sdk/octokit";
+import { LOG_LEVEL, LogLevel, Logs } from "@ubiquity-os/ubiquity-os-logger";
+import pkg from "../../package.json" with { type: "json" };
+import { createCronDatabase } from "./database-handler";
+
+async function main() {
+  const logger = new Logs((process.env.LOG_LEVEL as LogLevel) ?? LOG_LEVEL.INFO);
+  const octokit = new Octokit({
```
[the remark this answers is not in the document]
gentlementlegen: 405ce8b92d9eb78cd23638147aecc3503f4ac6a5

src/plugin.ts
```diff
@@ -54,9 +56,14 @@ export async function runPlugin(context: Context) {
...
-        await issueMatching(context as Context<"issues.edited">);
-        return await issueChecker(context as Context<"issues.edited">);
+        if (isPluginEdit(context as Context<"issues.edited">)) {
+          logger.info("Plugin edit detected, will run issue matching and checker.");
+          await issueMatching(context as Context<"issues.edited">);
+          await issueChecker(context as Context<"issues.edited">);
+        } else {
+          await updateIssue(context as Context<"issues.edited">);
+        }
+        return;
```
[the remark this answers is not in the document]
gentlementlegen: @shiv810 so this is safe to remove, as well as the config timeout then?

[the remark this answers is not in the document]
gentlementlegen: 053ef55 and forgot to rename it my bad. also seems that it broke the tests, fixing

tests/main.test.ts
```diff
@@ -37,6 +39,9 @@ describe("Plugin tests", () => {
   beforeEach(async () => {
     drop(db);
     await setupTests();
+    await mock.module("../src/helpers/plugin-utils", () => ({
+      isPluginEdit: jest.fn(() => true),
```
whilefoo, line 43: you're now importing both jest and bun, I wonder how it even works :D

src/plugin.ts
```diff
@@ -50,13 +52,16 @@ export async function runPlugin(context: Context) {
   } else if (isIssueEvent(context)) {
     switch (eventName) {
       case "issues.opened":
-        await addIssue(context as Context<"issues.opened">);
-        await issueMatching(context as Context<"issues.opened">);
-        return await issueChecker(context as Context<"issues.opened">);
+        return await addIssue(context as Context<"issues.opened">);
```
whilefoo, line 55: we could keep `issueMatching` in `issues.opened` because it makes a new comment and doesn't interfere

tests/main.test.ts
gentlementlegen: I think that bun is just capable to run `jest`, when you started migrating the tests within `text-conversation-rewards` didn't you say that it "just worked"? Imports were still probably from `@jest/global` so technically jest within bun. Still you are right I should not mix both I'll try to move them all to bun!

gentlementlegen: From the docs:
> There's often no need for code changes.
Bun internally re-writes imports from @jest/globals to use the bun:test equivalents.
If you're relying on Jest to inject test, expect, etc. as globals, Bun does that too.
But if you'd rather switch to the bun:test imports, you can do that too.

https://bun.com/guides/test/migrate-from-jest

whilefoo: ah interesting, yeah for `text-conversation-rewards` I switched imports to bun

src/plugin.ts
whilefoo: the issue matcher checks for existing comment and deletes it so there might be a bug there

whilefoo (commented):
I also get an [error](https://github.com/ubiquity-whilefoo/text-vector-embeddings/actions/runs/16734318268/job/47369625758#step:4:1370) when it's trying to update cron state:
```
Error: [@octokit/auth-app] appId option is required
    at authStrategy (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/node_modules/@octokit/auth-app/dist-node/index.js:430:1)
    at new super (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/node_modules/@octokit/core/dist-src/index.js:100:1)
    at new t (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/node_modules/@octokit/core/dist-src/index.js:19:1)
    at updateCronState (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/src/cron/workflow.ts:12:1)
    at runPlugin (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/src/plugin.ts:77:1)
    at context (ext:core/01_core.js:179:7)
    at result (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/src/main.ts:5:1)
    at createActionsPlugin (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/node_modules/@ubiquity-os/plugin-sdk/dist/index.mjs:413:1)
	⚠ Error updating cron workflow state
	⚠ {
	    "e": {},
	    "caller": "updateCronState"
	  }
	⚠   ↳  runPlugin (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/src/plugin.ts:77:1)
	    ↳  result (/home/runner/work/text-vector-embeddings/text-vector-embeddings/dist/src/main.ts:5:1)
```
It seems `APP_*` envs are missing in compute workflow

src/handlers/add-issue.ts
```diff
@@ -1,4 +1,4 @@
-import { Context } from "../types";
+import { Context } from "../types/index";
 import { removeFootnotes } from "./issue-deduplication";
 
 export async function addIssue(context: Context<"issues.opened">) {
```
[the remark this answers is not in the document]
gentlementlegen: 6b89161 should take care of it

whilefoo (approved):
QA on issue opened:
https://github.com/ubiquity-whilefoo/text-vector-embeddings/actions/runs/16775151364
https://github.com/ubiquity-whilefoo/text-vector-embeddings/actions/runs/16775167628
https://github.com/ubiquity-whilefoo/text-vector-embeddings/actions/runs/16775175021

PR #109, opened 2025-08-15
Add image description according to the image content

### LINKED ISSUE #58: Convert Images to Descriptions
In [this example](https://github.com/ubiquity/work.ubq.fi/issues/180), the text description without understanding the contents of the image is meaningless. For task matchmaking, a simple solution could be to replace embedded images with plain text descriptions of the image contents within the database when generating vector embeddings.

I'm unsure what the most cost efficient and highest performing tool would be. 

I suppose we should do the same for comments as well.

gentlementlegen:
Resolves #58

QA:
https://github.com/Meniole/text-vector-embeddings/actions/runs/17047900466/job/48328268765#step:4:140


src/types/plugin-input.ts
```diff
@@ -20,6 +21,22 @@ export const pluginSettingsSchema = T.Object(
       T.Number({ default: 0, description: "If set to a value greater than 0, the bot will always recommend contributors, regardless of the similarity score." })
     ),
     demoFlag: T.Boolean({ default: false, description: "When true, disables storing issues and comments in the database." }),
+    llm: T.Object(
+      {
+        model: T.String({
+          default: "google/gemini-2.5-flash-lite",
```
0x4007, line 27: I wonder if it makes sense to use the `openrouter/auto` model for the default?

https://openrouter.ai/openrouter/auto

gentlementlegen: @0x4007 yes sure we can, however there is no free / cheap model (mostly more performant, expensive ones) is that ok?

gentlementlegen: I guess it's better to have a cheap model in case we forget to configure it, or if we forget to mock it during tests to avoid spending a lot of tokens by mistake 😄

